PR #8, opened 2024-01-19
dep: lock alloy deps

rkrasiuk:
## Description

Lock alloy deps to last commit hash.


onbjerg (commented):
if we want this we need to lock at a similar hash in reth or the compiler will complain

we also need to update to a new commit since there have been changes in rpc types

onbjerg (approved):

PR #9, opened 2024-01-19
feat: use inspector db directly in js

### LINKED ISSUE #6: Get rid of js db service
this was initially added because we weren't able to share the db instance with the js context

this should be possible now via the guarded refs

mattsse:
Closes #6

This gets rid of the additional JS service requirement by transmuting the db reference and guarding it against garbage collection.

This applies a similar transmute hack as `std::thread::scoped` by transmuting the lifetime bound of the DB via a trait object.

The DB wrappers are nullable, once the guard is dropped the actual garbage collected DB reference is also removed

### DIFF
--- a/src/tracing/js/bindings.rs
+++ b/src/tracing/js/bindings.rs
@@ -738,7 +738,7 @@ impl EvmDbRef {
         // SAFETY:
         //
         // boa requires 'static lifetime for all objects.
-        // As mention in the `Safety` section of [GuardedNullableGc] the caller of this function
+        // As mentioned in the `Safety` section of [GuardedNullableGc] the caller of this function
         // needs to guarantee that the passed-in lifetime is sufficiently long for the lifetime of
         // the guard.
         let db = JsDb(db);
--- a/src/tracing/js/bindings.rs
+++ b/src/tracing/js/bindings.rs
@@ -6,7 +6,7 @@ use crate::tracing::{
             address_to_buf, bytes_to_address, bytes_to_hash, from_buf, to_bigint, to_buf,
             to_buf_value,
         },
-        JsDbRequest, TransactionContext,
+        TransactionContext,
     },
     types::CallKind,
 };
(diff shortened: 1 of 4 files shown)


src/tracing/js/bindings.rs
```diff
@@ -871,11 +913,61 @@ unsafe impl Trace for EvmDbRef {
     empty_trace!();
 }
 
+impl Clone for EvmDbRef {
```
DaniPopes, line 916: Derivable

src/tracing/js/bindings.rs
```diff
@@ -70,18 +70,29 @@ macro_rules! js_value_capture_getter {
...
+    inner: Rc<RefCell<Option<Guarded<'static, Val>>>>,
 }
 
-impl<Val: 'static> GuardedNullableGcRef<Val> {
-    /// Creates a garbage collectible reference to the given reference.
+impl<Val: 'static> GuardedNullableGc<Val> {
+    /// Creates a garbage collectible value to the given reference.
+    ///
+    /// SAFETY; the caller must ensure that the guard is dropped before the value is dropped.
+    pub(crate) fn r#ref(val: &Val) -> (Self, GcGuard<'_, Val>) {
```
DaniPopes, line 82: new_ instead of r#?

src/tracing/js/bindings.rs
```diff
@@ -680,39 +722,51 @@ impl EvmContext {
...
+    where
+        DB: DatabaseRef,
+        DB::Error: std::fmt::Display,
+    {
+        let (state, state_guard) = StateRef::new(state);
+
+        // SAFETY:
+        //
+        // boa requires 'static lifetime for all objects.
+        // As mention in the `Safety` section of [GuardedNullableGc] the caller of this function
```
DaniPopes, line 741:
```suggestion
        // As mentioned in the `Safety` section of [GuardedNullableGc] the caller of this function
```

src/tracing/js/bindings.rs
```diff
@@ -94,25 +105,35 @@ impl<Val: 'static> GuardedNullableGcRef<Val> {
...
+}
+
+/// Guard the inner value, once this value is dropped the inner value is also removed.
 ///
-/// This type guarantees that it never outlives the wrapped reference.
+/// This type guarantees that it never outlives the wrapped value.
 #[derive(Debug)]
-pub(crate) struct RefGuard<'a, Val> {
-    inner: Rc<RefCell<Option<&'a Val>>>,
+pub(crate) struct GcGuard<'a, Val> {
```
DaniPopes, line 132: #[must_use]?

DaniPopes (changes requested):

src/tracing/js/bindings.rs
mattsse: we already have new, this calls `Self::new` with a ref value, the function match the enum variants

DaniPopes (approved):
lgtm, weird Gc stuff is fine as long as it's private

src/tracing/js/bindings.rs
DaniPopes: Oh I meant new_ref instead of #ref

PR #15, opened 2024-01-28
fix: use original value in prestate tracer

### LINKED ISSUE #13: Geth PreState tracer has problems
It looks like PreStateTracer with diff_mode = Some(false) or None that means it collects initial state for the transaction returns intermediate value. If there's SSTORE(key, 1) SSTORE(key, 0) it returns 1, maybe it depends on SLOAD(key) used in between. 

I guess the problem is, when you setup touched cells here : https://github.com/paradigmxyz/evm-inspectors/blob/5ee90076f88e64a059e4c0c268929cf9ec91ed6d/src/tracing/builder/geth.rs#L205)

you should overwrite values even if they are already in the storage with 
`entry.insert(slot.previous_or_original_value.into());`
here
https://github.com/paradigmxyz/evm-inspectors/blob/5ee90076f88e64a059e4c0c268929cf9ec91ed6d/src/tracing/builder/geth.rs#L241

I found the problem with reth 0.16-alpha from todays main, checked my code with infura, and there's a difference indeed. With state from Infura transaction is commited well to REMV, however with Reth it looks like UniswapV3 lock is locked from the beginning.

mattsse:
Closes #13

~~This could potentially be simplified if we can simply use the result's statediff if this has the initial value for modified slots and not just the previous~~

because the prestate tracer should only include changed slots:

https://github.com/ethereum/go-ethereum/blob/bc0b87ca196f92e5af49bd33cc190ef0ec32b197/eth/tracers/native/prestate.go#L216-L224

simplifies prestate result by only checking changed accounts and using the original value

cc @rakita

### DIFF
--- a/src/tracing/builder/geth.rs
+++ b/src/tracing/builder/geth.rs
@@ -202,7 +202,8 @@ impl GethTraceBuilder {
                     Entry::Occupied(entry) => entry.into_mut(),
                 };
 
-                for (key, value) in node.touched_slots() {
+                // TODO is this even necessary if we only need the changed slots? because all of them should be included in the `account_diffs`
+                for (key, value) in node.touched_slots_original_values() {
                     match acc_state.storage.entry(key.into()) {
                         Entry::Vacant(entry) => {
                             entry.insert(value.into());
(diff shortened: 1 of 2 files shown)


onbjerg (approved):
lgtm, nice cleanup

PR #16, opened 2024-01-30
change: Expose additional fields

jasalper:
Exposes a few extra functions in CallTraceNode so that they're public to users of the crate rather than just internally. (In case someone wants to make parity or geth-like traces


mattsse (approved):
reasonable

PR #24, opened 2024-02-09
feat: add op counter 

DoTheBestToGetTheBest:
related with https://github.com/paradigmxyz/evm-inspectors/issues/20


mattsse (changes requested):
@sslivkoff what should `opcode_gas`

the total gas cost per opcode * count?

DoTheBestToGetTheBest:
@mattsse does it look good to you ?

DoTheBestToGetTheBest:
> the total gas cost per opcode * count?


I was thinking of mapping each opcode to its constant gas value. Then, when the final opcode result is mapped, I would simply multiply the constant gas by the number of repetitions of that opcode.  but  this implementation seems inefficient, doesn't it ?

DoTheBestToGetTheBest:
or it may be some funciton that can calculate the gas? 

> @sslivkoff what should `opcode_gas`
> 
> the total gas cost per opcode * count?

i tried to use the look up table from revm but it doesn't work it return for some opcode gas 0 and for some good value and for others false value


```rust
   fn step(&mut self, interp: &mut Interpreter, _context: &mut EvmContext<DB>) {
   
        let opcode_val = interp.current_opcode();
        let opcode =
            OpCode::new(opcode_val).unwrap_or_else(|| unsafe { OpCode::new_unchecked(opcode_val) });

  
        *self.opcode_counts.entry(opcode).or_insert(0) += 1;

        let spec_id = revm::primitives::specification::SpecId::SHANGHAI;
        let gas_info =
            revm::interpreter::instructions::opcode::spec_opcode_gas(spec_id)[opcode_val as usize];

        let opcode_gas = gas_info.get_gas() as u64;

        *self.opcode_gas.entry(opcode).or_insert(0) += opcode_gas;
    }
}```

**Comment by sslivkoff** [comment]
> @sslivkoff what should `opcode_gas`
> 
> the total gas cost per opcode * count?

afaui the same op-code can cost different amounts of gas depending on context and depending on input arguments

so ideally it would be the sum of each opcode ocurrence but this would not be a simple multiplication

**Comment by sslivkoff** [comment]
there's also a bit of complication on contract calls returning whatever amount of spare unused gas they do not use

ideally want a situation where the sum of opcode gas entries equals the total amount of gas used by the transaction

might need to do some interesting nested gas accounting to make this happen given the various types of gas refunds

File: src/counter_op.rs [hunk 1]
```diff
@@ -15,19 +16,21 @@ impl OpcodeCounterInspector {
...
+impl<DB> revm::Inspector<DB> for OpcodeCounterInspector
 where
-    DB: Database,
+    DB: revm::Database,
 {
-    fn step(&mut self, interp: &mut Interpreter, _context: &mut EvmContext<DB>) {
+    fn step(&mut self, interp: &mut Interpreter, _context: &mut revm::EvmContext<DB>) {
         let opcode_value = interp.current_opcode();
+        let opcode = OpCode::new(opcode_value)
+            .unwrap_or_else(|| unsafe { OpCode::new_unchecked(opcode_value) });
```
mattsse, line 31: it's possible that this is an invalid opcode, so we can't use unsafe here and instead need to try and only if it's a valid opcode add it to the set

mattsse (changes requested):

src/counter_op.rs
DoTheBestToGetTheBest: > it's possible that this is an invalid opcode, so we can't use unsafe here and instead need to try and only if it's a valid opcode add it to the set

Thanks for your review Mat! Changed this :)

DoTheBestToGetTheBest: if you don't mind i need some help to count the gas of the opcode, since [sslivkoff] doesn't want the multiplication i dunno how i would track this

DoTheBestToGetTheBest: cc @mattsse  its possible to have review of this pls

src/counter_op.rs
```diff
@@ -1,39 +1,59 @@
...
+        if let Some(opcode) = OpCode::new(opcode_val) {
+            *self.opcode_counts.entry(opcode).or_insert(0) += 1;
 
-        let count = self.opcode_counts.entry(opcode).or_insert(0);
-        *count += 1;
+            
+            // this should be upgraded to new release
+            // when new ethereum version going to be updated in 2024/03
+            let gas_info = revm::interpreter::instructions::opcode::spec_opcode_gas(
+                revm::primitives::specification::SpecId::SHANGHAI,
```
mattsse, line 45: the specid we can get from the context `context.spec_id()`

mattsse (changes requested):

src/counter_op.rs
mattsse: thanks, almost done

mattsse:
let's also make the fields pub so they can be consumed easily

DoTheBestToGetTheBest:
> let's also make the fields pub so they can be consumed easily

hm they are already public,

src/counter_op.rs
DoTheBestToGetTheBest: > the specid we can get from the context `context.spec_id()`

thanks, didn't know about this (:

DoTheBestToGetTheBest:
@mattsse does it look goods to you ? workflow is failing but doesn't seem related with this pr ?

src/lib.rs
```diff
@@ -5,20 +5,32 @@
...
+)]
+#![cfg_attr(not(test), warn(unused_crate_dependencies))]
+#![deny(unused_must_use, rust_2018_idioms)]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
 
 /// An inspector implementation for an EIP2930 Accesslist
 pub mod access_list;
 
+/// implementation of an opcode counter for the EVM.
+pub mod counter_op;
```
mattsse, line 30:
```suggestion
pub mod opcode;
```

src/counter_op.rs
```diff
@@ -0,0 +1,87 @@
+use revm::{
+    interpreter::{opcode::OpCode, Interpreter},
+    Database, EvmContext, Inspector,
+};
+use std::collections::HashMap;
+
+/// An Inspector that counts opcodes and measures gas usage per opcode.
+#[derive(Default, Debug)]
```
mattsse, line 8:
```suggestion
#[derive(Default, Debug, Clone)]
```

mattsse (changes requested):
I few nits,

this could also use a helper function that returns an iterator over all opcodes and their count+gas (opcode, (u64,u64))

DoTheBestToGetTheBest:
> I few nits,
> 
> this could also use a helper function that returns an iterator over all opcodes and their count+gas (opcode, (u64,u64))

thanks, i hope this time look good to you :)

mattsse (approved):
lgtm

PR #45, opened 2024-02-17
Fix: track dynamic gas used in opcode tracking gas

DoTheBestToGetTheBest:
related : https://github.com/paradigmxyz/evm-inspectors/issues/40

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -1,22 +1,27 @@
 use revm::{
+    inspectors::GasInspector,
     interpreter::{opcode::OpCode, Interpreter},
     Database, EvmContext, Inspector,
 };
 use std::collections::HashMap;
-
 /// An Inspector that counts opcodes and measures gas usage per opcode.
-#[derive(Clone, Debug, Default)]
+#[derive(Default, Debug)]
 pub struct OpcodeCounterInspector {
     /// Map of opcode counts per transaction.
     pub opcode_counts: HashMap<OpCode, u64>,
     /// Map of total gas used per opcode.
     pub opcode_gas: HashMap<OpCode, u64>,
+    gas_inspector: GasInspector,
 }
 
 impl OpcodeCounterInspector {
     /// Creates a new instance of the inspector.
     pub fn new() -> Self {
-        Self { opcode_counts: HashMap::new(), opcode_gas: HashMap::new() }
+        OpcodeCounterInspector {
+            opcode_counts: HashMap::new(),
+            opcode_gas: HashMap::new(),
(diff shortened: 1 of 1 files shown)


src/opcode.rs
```diff
@@ -42,18 +47,23 @@ impl<DB> Inspector<DB> for OpcodeCounterInspector
 where
     DB: Database,
 {
-    fn step(&mut self, interp: &mut Interpreter, _context: &mut EvmContext<DB>) {
+    fn step(&mut self, interp: &mut Interpreter, context: &mut EvmContext<DB>) {
```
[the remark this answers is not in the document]
DoTheBestToGetTheBest: > we now need to implement all the other Inspector functions that the gasinspector also implements

thanks for your review boss!

src/opcode.rs
```diff
@@ -42,20 +47,67 @@ impl<DB> Inspector<DB> for OpcodeCounterInspector
...
+        self.initialize_interp(interp, context);
```
mattsse, line 51: here we need to delegate to gas inspector

src/opcode.rs
```diff
@@ -42,20 +47,67 @@ impl<DB> Inspector<DB> for OpcodeCounterInspector
...
+        self.step_end(interp, context);
+        let gas_table = revm::interpreter::instructions::opcode::spec_opcode_gas(context.spec_id());
+        let opcode_gas_info = gas_table[opcode_value as usize];
 
-            let opcode_gas_cost = opcode_gas_info.get_gas() as u64;
+        let opcode_gas_cost = opcode_gas_info.get_gas() as u64;
 
-            *self.opcode_gas.entry(opcode).or_insert(0) += opcode_gas_cost;
+        if let Some(opcode) = OpCode::new(opcode_value) {
+            let gas_used = self.gas_inspector.last_gas_cost();
```
mattsse, line 65: this is a bit tricky because gas cost track cost of the entire call so far

we need to track gas at the beginning of the step and on step end we can get the diff by comparing `.remaining`

see also TracingInspector for reference

mattsse (changes requested):
almost there

the gas inspector is a bit tricky to use for these things

src/opcode.rs
DoTheBestToGetTheBest: > this is a bit tricky because gas cost track cost of the entire call so far
> 
> we need to track gas at the beginning of the step and on step end we can get the diff by comparing `.remaining`
> 
> see also TracingInspector for reference

hey thank you for the review, i'm not sure i understand this

mattsse (approved):
thanks for this!

PR #47, opened 2024-02-19
feat: add feature-gated Serde implementations

DaniPopes:
Ref https://github.com/foundry-rs/foundry/issues/6922


mattsse (approved):
lgtm

mental note that we have to enable this on reth when we bump

PR #52, opened 2024-03-03
fix(tracing): collect logs only if call succeeded in geth `callTracer`

ArtificialPB:
Logs should only be collected for calls where it and all of its parents succeeded. 

This logic is ported from geth `callTracer`: https://github.com/ethereum/go-ethereum/blob/master/eth/tracers/native/call.go#L278-L289


src/tracing/builder/geth.rs
```diff
@@ -169,6 +172,18 @@ impl GethTraceBuilder {
...
+    /// Returns true if the given trace or any of its parents failed.
+    fn call_or_parent_failed(&self, node: &CallTraceNode) -> bool {
+        if node.trace.is_error() {
+            return true;
+        }
+
+        if let Some(parent) = node.parent {
+            return self.call_or_parent_failed(&self.nodes[parent]);
+        }
+        false
```
mattsse, line 184: can we solve this iteratively by looping instead?

mattsse (changes requested):
nice, 

I'd like to use `while let Some` here

src/tracing/builder/geth.rs
ArtificialPB: done

mattsse (approved):
lgtm, ty!

DaniPopes (approved):

PR #53, opened 2024-03-04
feat: bump alloy rpc types rev

fgimenez:
Updates `alloy-rpc-types` and `alloy-rpc-trace-types` to `785c667813a6c76794044b943df58fc6e397734d`, which contains https://github.com/alloy-rs/alloy/pull/245


mattsse (approved):
we should really get rid of this and start publishing rpc types

PR #11, opened 2024-01-22
feat: migrate to new inspector API

mattsse:
Closes #2

Blocked by reth migration

and 

https://github.com/bluealloy/revm/issues/1001

PR #56, opened 2024-03-06
chore: remove unused code var

mattsse:
this is a leftover of #11 

@rakita 

the `output` of `CreateOutcome` is the runtime bytecode, right?

https://github.com/paradigmxyz/evm-inspectors/blob/05352cd16db13fd89e9a974487b456e3f1498c69/src/tracing/types.rs#L48-L50


DaniPopes (approved):

DaniPopes:
That is right

PR #57, opened 2024-03-06
feat(tracing): implement muxTracer

ArtificialPB:
Adds support for `muxTracer`. Depends on changes from https://github.com/alloy-rs/alloy/pull/252

Once both PR's are good to go I'll open a PR also for `reth`.

### DIFF
--- a/src/tracing/mux.rs
+++ b/src/tracing/mux.rs
@@ -14,24 +14,23 @@ use std::collections::HashMap;
 use thiserror::Error;
 
 /// Mux tracing inspector that runs and collects results of multiple inspectors at once.
+///
+/// Contains a list of tracer types with its inspectors.
 #[derive(Clone, Debug)]
-pub struct MuxInspector {
-    /// The list of tracer types with its inspectors.
-    inspectors: Vec<(GethDebugBuiltInTracerType, DelegatingInspector)>,
-}
+pub struct MuxInspector(Vec<(GethDebugBuiltInTracerType, DelegatingInspector)>);
 
 impl MuxInspector {
     /// Try creating a new instance of [MuxInspector] from the given [MuxConfig].
     pub fn try_from_config(config: MuxConfig) -> Result<MuxInspector, Error> {
         let inspectors = config
-            .tracers
+            .0
             .into_iter()
             .map(|(tracer_type, tracer_config)| {
                 DelegatingInspector::try_from_config(tracer_type, tracer_config.clone())
             })
             .collect::<Result<Vec<_>, _>>()?;
 
-        Ok(MuxInspector { inspectors })
+        Ok(MuxInspector(inspectors))
     }
 
     /// Try converting this [MuxInspector] into a [MuxFrame].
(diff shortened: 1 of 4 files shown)


src/tracing/mux.rs
```diff
@@ -0,0 +1,408 @@
...
+                inspector.create_end(context, inputs, outcome)
+            }
+        }
+    }
+
+    #[inline]
+    fn selfdestruct<DB: Database>(&mut self, contract: Address, target: Address, value: U256) {
+        match self {
+            MultiInspector::FourByteInspector(inspector) => {
+                <FourByteInspector as Inspector<DB>>::selfdestruct(
```
mattsse, line 354: I faced the same issue with the selfdestruct call but couldn't figure out why this is necessary here

src/tracing/mux.rs
```diff
@@ -0,0 +1,388 @@
...
+    #[inline]
+    fn call(
+        &mut self,
+        context: &mut EvmContext<DB>,
+        inputs: &mut CallInputs,
+    ) -> Option<CallOutcome> {
+        for (_, inspector) in &mut self.inspectors {
+            if let Some(outcome) = inspector.call(context, inputs) {
+                return Some(outcome);
+            }
```
mattsse, line 108: I believe the `TracingInspector` does not override the response so we can ignore this some check

src/tracing/mux.rs
```diff
@@ -0,0 +1,388 @@
...
+    #[inline]
+    fn create(
+        &mut self,
+        context: &mut EvmContext<DB>,
+        inputs: &mut CreateInputs,
+    ) -> Option<CreateOutcome> {
+        for (_, inspector) in &mut self.inspectors {
+            if let Some(outcome) = inspector.create(context, inputs) {
+                return Some(outcome);
+            }
```
mattsse, line 138: same

src/tracing/mux.rs
```diff
@@ -0,0 +1,388 @@
...
+/// Mux tracing inspector that runs and collects results of multiple inspectors at once.
+#[derive(Clone, Debug)]
+pub struct MuxInspector {
+    /// The list of tracer types with its inspectors.
+    inspectors: Vec<(GethDebugBuiltInTracerType, MultiInspector)>,
+}
+
+impl MuxInspector {
+    /// Try creating a new instance of [MuxInspector] from the given [MuxConfig].
+    pub fn try_from_config(config: MuxConfig) -> Result<MuxInspector, Error> {
```
mattsse, line 25: hmm, maybe we should actually make MultiInspector pub so this type can be configured easily manually without needing MuxConfig with an additional new function

```
pub new(impl IntoIter<MultiInspector>) -> Self
```

but not required for this pr

mattsse (changes requested):
this looks great,
a few nits and suggestions

src/tracing/mux.rs
ArtificialPB: I think it's because the `FourByteInspector` implements the `Inspector<DB>` trait, where `selfdestruct` function has no reference to the `DB` type, and compiler can't infer the `Inspector` trait type

src/tracing/mux.rs
```diff
@@ -0,0 +1,408 @@
...
+    fn selfdestruct(&mut self, contract: Address, target: Address, value: U256) {
+        for (_, inspector) in &mut self.inspectors {
+            inspector.selfdestruct::<DB>(contract, target, value);
+        }
+    }
+}
+
+/// An inspector that can delegate to multiple inspector types.
+#[derive(Clone, Debug)]
+enum MultiInspector {
```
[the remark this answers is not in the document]
ArtificialPB: I've renamed it to `DelegatingInspector`, seems like a better name

mattsse (approved):
lgtm

PR #60, opened 2024-03-08
feat: more geth tracer config functions

mattsse:
blocked by https://github.com/bluealloy/revm/pull/1174

the `GethDefaultTracingOptions` dont have any effect on the call and prestate tracer

this adds special functions for those tracers.

The prestate tracer can be simplified entirely and doesn't even have to be a tracer I think, but need to handle that separately


DaniPopes:
Test failures due to breaking change in revm release

DaniPopes (approved):

PR #61, opened 2024-03-08
bump: revm

rakita:
Opening PR just to check if CI passes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,11 +25,11 @@ alloy-sol-types = "0.6"
 alloy-primitives = "0.6"
 alloy-rpc-types = { git = "https://github.com/alloy-rs/alloy", rev = "64d0352" }
 alloy-rpc-trace-types = { git = "https://github.com/alloy-rs/alloy", rev = "64d0352" }
-revm = { version = "6.0", default-features = false, features = ["std"] }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -45,3 +45,9 @@ expect-test = "1.4"
 [features]
 serde = ["dep:serde", "revm/serde"]
 js-tracer = ["dep:boa_engine", "dep:boa_gc"]
+
+[patch.crates-io]
+revm = { git = "https://github.com/bluealloy/revm", branch = "inner_evm_context" }
+revm-precompile = { git = "https://github.com/bluealloy/revm", branch = "inner_evm_context" }
+revm-primitives = { git = "https://github.com/bluealloy/revm", branch = "inner_evm_context" }
+revm-interpreter = { git = "https://github.com/bluealloy/revm", branch = "inner_evm_context" }
```
gakonst, line 53: vote to just use the crates with the `branch` tag instead of the patch as it makes lib usage rly hard

mattsse (approved):

Cargo.toml
mattsse: hence the draft, will be updated after merge+release

rakita: This is just for a test, will publish revm before merging this

PR #63, opened 2024-03-09
chore: add Inspector::fuse

mattsse:
prep for making Inspectors reusable for multiple runs

### DIFF
--- a/src/tracing/mod.rs
+++ b/src/tracing/mod.rs
@@ -93,7 +93,7 @@ impl TracingInspector {
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/tracing/mod.rs
```diff
@@ -93,6 +93,37 @@ impl TracingInspector {
         }
     }
 
+    /// Resets the inspector to it's initial state of [Self::new].
```
DaniPopes, line 96:
```suggestion
    /// Resets the inspector to its initial state of [Self::new].
```

DaniPopes (approved):

PR #71, opened 2024-03-12
fix: use JSON.stringify for JS result

### LINKED ISSUE #69: Native JS BigInt type for proper serialization
the Js Bigint type does not serialize as hex:

https://github.com/paradigmxyz/reth/issues/7083

we need to roll our own JsValue for this that behaves as BigInt

mattsse:
closes #69

closes https://github.com/paradigmxyz/reth/issues/7083

use builtin JSON.stringify when formatting the result, boa.to_json uses all properties

### DIFF
--- a/src/tracing/js/bindings.rs
+++ b/src/tracing/js/bindings.rs
@@ -969,7 +969,9 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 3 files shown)


DaniPopes (approved):
Absolutely disgusting

PR #75, opened 2024-03-21
feat: write instruction result when displaying call traces

DaniPopes:
See https://github.com/foundry-rs/foundry/pull/7465


mattsse (approved):
that's much better

PR #88, opened 2024-04-07
chore: create unknown opcodes as unchecked

mattsse:
this could be an opcode that is manually added to the evm

### DIFF
--- a/src/tracing/js/bindings.rs
+++ b/src/tracing/js/bindings.rs
@@ -382,16 +382,11 @@ impl OpObj {
         let to_string = FunctionObjectBuilder::new(
             context.realm(),
             NativeFunction::from_copy_closure(move |_this, _args, _ctx| {
-                let op = OpCode::new(value)
-                    .or_else(|| {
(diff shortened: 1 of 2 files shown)


src/tracing/js/bindings.rs
```diff
@@ -382,16 +382,11 @@ impl OpObj {
...
-                    .or_else(|| {
-                        // if the opcode is invalid, we'll use the invalid opcode to represent it
-                        // because this is invoked before the opcode is
-                        // executed, the evm will eventually return a `Halt`
-                        // with invalid/unknown opcode as result
-                        let invalid_opcode = 0xfe;
-                        OpCode::new(invalid_opcode)
-                    })
-                    .expect("is valid opcode;");
+                let op = OpCode::new(value).unwrap_or_else(|| {
```
DaniPopes, line 385: Why unwrap_or_else? Also cc https://github.com/bluealloy/revm/pull/1215
This is what I meant when removing unsafe from `new_unchecked`.

mattsse: good point, we can just do new_unched here

PR #111, opened 2024-05-07
fix: selfdestructs once and for all

DaniPopes:
In >=Cancun `selfdestruct` inspector function is not called, but `SELFDESTRUCT` is still returned as instruction result.

Found while debugging https://github.com/foundry-rs/foundry/pull/7880


DaniPopes:
I realize that theres no way to get the refund address in cancun because the inspector isnt called unless im missing something

mattsse (approved):
selfdestructs 👎

DaniPopes:
yes but the value transfer still happens, it's just that we can't get the value of it

PR #115, opened 2024-05-11
Fix index out of bound when resetting TracingInspector

Ebolon:
Fixes out of bound error when using `fuse()` in `TracingInspector`.

Error before:
```
index out of bounds: the len is 0 but the index is 0
panicked at src/tracing/arena.rs:54:31:
```

Cause: `CallTraceArena` is initialized using `default()` but when calling `clear()` the default root node is removed.


mattsse (approved):
nice find.

I kept using clear, but pushed the empty node.

PR #119, opened 2024-05-12
wip:fix: included all accessed changed slots

mattsse:
closes https://github.com/paradigmxyz/reth/issues/8202

we need to manually track reverted changed slots in case they're handled gracefully


rakita:
I would close this PR in the favor of this change: https://github.com/bluealloy/revm/pull/1437

This would preserve any account that got loaded inside revm and would make it easier to construct full access list

mattsse:
nice!

PR #142, opened 2024-06-13
perf: remove GasInspector from tracer, optimize step*

DaniPopes:
Only `gas_inspector.gas_remaining()` was used, which is equivalent to `interp.gas.remaining()` in the context of `step` and `step_end`.

Additionally, mark the internal `on_step` functions as `#[cold]`, as usually these are the cold path for heavy tracing operations or debugging, rather than the common path.


mattsse (approved):
lgtm

PR #148, opened 2024-06-17
deps: bump revm

shekhirin:


mattsse (approved):
if we intend to bump in reth we likely need to transition to alloy via crates as well

PR #152, opened 2024-06-24
feat: small updates for steps tracing

klkvr:
- Adds `record_returndata_snapshots` flag to config which enables snapshots of `interpreter.return_data_buffer`
- Adds `record_opcodes_filter` parameter which allows to only record specific opcodes. ref https://github.com/foundry-rs/foundry/pull/8222#issuecomment-2186811858
- Adds `gas_used` field for `CallTraceStep`

This should be enough to migrate foundry's debugger to using `TracingInspector` from here, I will open PR for this later today.

### DIFF
--- a/src/tracing/config.rs
+++ b/src/tracing/config.rs
@@ -2,7 +2,27 @@ use alloy_rpc_types::trace::{
     geth::{CallConfig, GethDefaultTracingOptions, PreStateConfig},
     parity::TraceType,
 };
+use revm::interpreter::OpCode;
 use std::collections::HashSet;
+
+/// 256 bits each marking whether an opcode should be included into steps trace or not.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct OpcodeFilter([bool; 256]);
+
+impl OpcodeFilter {
+    pub const fn new() -> Self {
+        Self([false; 256])
+    }
+
+    pub fn enabled(&self, op: &OpCode) -> bool {
+        self.0[op.get() as usize]
+    }
+
+    pub const fn enable(mut self, op: OpCode) -> Self {
+        self.0[op.get() as usize] = true;
+        self
+    }
+}
 
 /// Gives guidance to the [TracingInspector](crate::tracing::TracingInspector).
 ///
@@ -18,6 +38,11 @@ pub struct TracingInspectorConfig {
     pub record_stack_snapshots: StackSnapshotType,
     /// Whether to record state diffs.
     pub record_state_diff: bool,
+    /// Whether to record returndata buffer snapshots.
+    pub record_returndata_snapshots: bool,
+    /// Optional filter for opcodes to record. If provided, only steps with opcode in this set will
+    /// be recorded.
+    pub record_opcodes_filter: Option<OpcodeFilter>,
     /// Whether to ignore precompile calls.
     pub exclude_precompile_calls: bool,
     /// Whether to record logs
@@ -32,6 +57,8 @@ impl TracingInspectorConfig {
             record_memory_snapshots: true,
             record_stack_snapshots: StackSnapshotType::Full,
             record_state_diff: false,
+            record_returndata_snapshots: true,
+            record_opcodes_filter: None,
             exclude_precompile_calls: false,
             record_logs: true,
         }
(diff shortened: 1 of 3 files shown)


src/tracing/config.rs
```diff
@@ -2,8 +2,37 @@ use alloy_rpc_types::trace::{
     geth::{CallConfig, GethDefaultTracingOptions, PreStateConfig},
     parity::TraceType,
 };
+use revm::interpreter::OpCode;
 use std::collections::HashSet;
 
+/// 256 bits each marking whether an opcode should be included into steps trace or not.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct OpcodeFilter([bool; 256]);
```
DaniPopes, line 10: This is 256 bytes not bits, use `U256` to make it bits

src/tracing/config.rs
```diff
@@ -2,8 +2,37 @@ use alloy_rpc_types::trace::{
...
+}
+
+impl OpcodeFilter {
+    /// Returns a new [OpcodeFilter] that does not trace any opcodes.
+    pub const fn new() -> Self {
+        Self([false; 256])
+    }
+
+    /// Returns whether steps with given [OpCode] should be traced.
+    pub fn enabled(&self, op: &OpCode) -> bool {
```
DaniPopes, line 25:
```suggestion
    pub fn is_enabled(&self, op: OpCode) -> bool {
```

DaniPopes (changes requested):

klkvr:
this conflicts with approved https://github.com/paradigmxyz/revm-inspectors/pull/84 as it relies on previous recorded step always containing memory during previous instruction. This is wouldn't be true if `record_opcodes_filter` is configured

DaniPopes:
We could just override the filter to say that we're tracing memory opcodes if we also want memory in that case

mattsse:
> We could just override the filter to say that we're tracing memory opcodes if we also want memory in that case

that makes sense

klkvr:
> We could just override the filter to say that we're tracing memory opcodes if we also want memory in that case

hmm, perhaps we should just fallback to old behavior in that case? if user provided filter to only collect `JUMP`s for example it might be unexpected that we'd collect 11 other opcodes as well potentially increasing memory usage

src/tracing/mod.rs
```diff
@@ -334,7 +334,17 @@ impl TracingInspector {
         let trace = &mut self.traces.arena[trace_idx];
 
         let step_idx = trace.trace.steps.len();
-        self.step_stack.push(StackStep { trace_idx, step_idx });
+        // we always want an OpCode, even it is unknown because it could be an additional opcode
+        // that not a known constant
+        let op = unsafe { OpCode::new_unchecked(interp.current_opcode()) };
+
+        let record = self.config.record_opcodes_filter.as_ref().map_or(true, |f| f.is_enabled(&op));
```
mattsse, line 341: could we make this a function instead?

src/tracing/mod.rs
```diff
@@ -562,4 +586,5 @@ where
 struct StackStep {
     trace_idx: usize,
     step_idx: usize,
+    record: bool,
```
mattsse, line 589: needs docs what for what this tracks

src/tracing/types.rs
```diff
@@ -506,10 +506,14 @@ pub struct CallTraceStep {
     pub memory: RecordedMemory,
     /// Size of memory at the beginning of the step
     pub memory_size: usize,
+    /// Returndata before step execution
+    pub returndata: Bytes,
```
mattsse, line 510: I do remember we had this at some point

mattsse (changes requested):
if I'm reading this correctly, then this does not change the existing recording logic and only adds returndata recording based on opcode filter?

the overhead is negligible here

we can do @DaniPopes suggestion separately I believe

mattsse (approved):
lgtm, pending @DaniPopes

klkvr:
> if I'm reading this correctly, then this does not change the existing recording logic and only adds returndata recording based on opcode filter?

opcode filter and returndata are separate configs

returndata is recorded in the same way as memory: if `record_returndata_snapshots` is true, we always save snapshot

and opcode filter is just a way to make tracer record only specific steps to optimize memory usage if we're only interested in a small subset of steps. e.g. this is the case for internal fns tracking in foundry where we only want `JUMP`s and `JUMPDEST`s along with their stack and memory snapshots

DaniPopes (approved):

PR #153, opened 2024-06-26
fix: track selfdestruct transferred value separately

mattsse:
in https://github.com/paradigmxyz/revm-inspectors/commit/ee6f6be412c61a494472468c37bc7b8bc292b458 we updated

`node.trace.value = value;` with the selfdestruct value transfer.

this results in incorrect traces for original call: https://github.com/paradigmxyz/reth/issues/9124

This tracks the transferred value in a separate field similar to refund address.


DaniPopes (approved):
lgmt

PR #156, opened 2024-06-30
feat(EOF): Enable inspector calls

rakita:
Bumps revm to newest release and adds support for EOF Inspector functions


src/tracing/mod.rs
```diff
@@ -589,6 +590,37 @@ where
         None
     }
 
+    fn eofcreate(
+        &mut self,
+        context: &mut EvmContext<DB>,
+        inputs: &mut EOFCreateInputs,
+    ) -> Option<CreateOutcome> {
+        let _ = context.load_account(inputs.caller);
+        let nonce = context.journaled_state.account(inputs.caller).info.nonce;
```
mattsse, line 599: looks like this is only required for the else case, can we do this with if let else instead?

mattsse (changes requested):
I think we also need to bump the msrv in the ci?

mattsse (approved):

PR #157, opened 2024-07-01
feat: add decoding for individual trace steps

klkvr:
ref https://github.com/foundry-rs/foundry/pull/8222
ref https://github.com/foundry-rs/foundry/issues/8198

Adds structs and extends `TraceWriter` to support formatting of decoded trace steps. Currently two decoding formats are supported:
- Internal calls. Similar to a decoded call trace, decoded internal function invocation which spans over multiple steps. Kept as decoded function name, inputs, outputs and index of the last step.
- Arbitrary strings. This might be useful for formatting decoded opcodes (e.g. adding `├─ [sload] <slot>` to trace. It might make sense to extend it to something more configurable once we start implementing this

### DIFF
--- a/src/tracing/types.rs
+++ b/src/tracing/types.rs
@@ -543,6 +543,7 @@ pub enum TraceMemberOrder {
 
 /// Represents a decoded internal function call.
 #[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct DecodedInternalCall {
     /// Name of the internal function.
     pub func_name: String,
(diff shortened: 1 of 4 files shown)


src/tracing/types.rs
```diff
@@ -541,6 +541,31 @@ pub enum TraceMemberOrder {
     Step(usize),
 }
 
+/// Represents a decoded internal function call.
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct DecodedInternalCall {
```
DaniPopes, line 547: maybe worth also adding an optional style field?

mattsse (approved):
lgtm,

pending @DaniPopes

src/tracing/types.rs
DaniPopes: Can be a follow-up where we add it to other structs as well

DaniPopes (approved):

PR #169, opened 2024-07-17
feat: add immediate bytes recording

klkvr:
before EOF those were only present for `PUSH*` opcodes, and they were trivial to obtain by looking at the top of the stack, now it is a bit more complex, so this would be useful

### DIFF
--- a/src/tracing/mod.rs
+++ b/src/tracing/mod.rs
@@ -417,6 +417,7 @@ impl TracingInspector {
         trace.trace.steps.push(CallTraceStep {
             depth: context.journaled_state.depth(),
             pc: interp.program_counter(),
+            code_section_idx: interp.function_stack.current_code_idx,
             op,
             contract: interp.contract.target_address,
             stack,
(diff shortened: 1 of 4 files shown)


src/tracing/mod.rs
```diff
@@ -404,6 +404,16 @@ impl TracingInspector {
         let gas_used =
             gas_used(context.spec_id(), interp.gas.spent(), interp.gas.refunded() as u64);
 
+        let immediate_bytes =
+            (self.config.record_immediate_bytes && op.info().immediate_size() > 0).then(|| {
+                let pc = interp.program_counter();
+                interp
+                    .bytecode
+                    .get(pc + 1..pc + 1 + op.info().immediate_size() as usize)
```
DaniPopes, line 412: note that this doesn't include the full `RJUMPV` immediate, which is calculated as `len = (imm[0] + 1) * 2`

DaniPopes, line 412: We can `slice` here to avoid allocating for each immediate

DaniPopes (changes requested):

DaniPopes (approved):

mattsse (approved):

Issue #162, opened 2024-07-12
parity selfdestruct traces are inserted out of order


mattsse:
https://github.com/paradigmxyz/reth/issues/9142

we don't record selfdestructs as individual trace nodes instead the node can have a selfdestruct state.

we compute the traceAddress correctly

               "traceAddress": [
                      2
                  ]

but they get inserted incorrectly, right after the top-level call:

https://github.com/paradigmxyz/evm-inspectors/blob/b9850ffe4d67aadc46cba5e3798bee459a01a560/src/tracing/builder/parity.rs#L238-L256

## TODO
* insert selfdestruct after the last subtrace

PR #170, opened 2024-07-23
Fix Self-destruct Disorder

ZzPoLariszZ:
The `suicide`/`selfdestruct` trace should not be immediately after the parent trace but after all of the parent's subtraces.

This can be fixed by ordering the `traces` using the attribute `trace_address` inside.

closes https://github.com/paradigmxyz/revm-inspectors/issues/162

### DIFF
--- a/src/tracing/builder/parity.rs
+++ b/src/tracing/builder/parity.rs
@@ -255,6 +255,8 @@ impl ParityTraceBuilder {
             }
         }
 
+        traces.sort_by(|a, b| a.trace_address.cmp(&b.trace_address));
+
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
this is just a much simpler solution than https://github.com/paradigmxyz/revm-inspectors/pull/163 right?

the sort is only necessary if we have a selfdestruct, right?

can we use a bool marker and do sorting conditionally?

mattsse (approved):
tysm, this is much simpler than what I previously attempted

PR #171, opened 2024-07-24
fix: gas and gasUsed in trace root only for ParityTrace

ZzPoLariszZ:
## Description

This pull request is Part 1/2 of fixing the bug where the `gas` and `gasUsed` fields in Parity Trace root are incorrect.

Part 2/2 paradigmxyz/reth#9761

## Related Issues and Pull Requests

- Follow: ethereum/go-ethereum#27029
- Improve: paradigmxyz/reth#3678 and paradigmxyz/reth#3719
- Fix: paradigmxyz/reth#9142 with #170 
- Update: paradigmxyz/reth#3782

## Problem

The `gas` and `gasUsed` fields in Geth Debug Trace root should be the gas limit and gas used for the entire transaction.

However, two fields in Parity Trace root should be the original ones.

### Reproducible Example

With the latest version Reth v1.0.3, using `trace_transaction()` to trace  
the transaction `0x03128677ee3a9623d20f3c677f423ccc592d126374bf32e331343dd1bdf38b61`

```
curl http://localhost:8545 \
  -X POST \
  -H "Content-Type: application/json" \
  --data '{"method":"trace_transaction","params":["0x03128677ee3a9623d20f3c677f423ccc592d126374bf32e331343dd1bdf38b61"],"id":1,"jsonrpc":"2.0"}'  
```

**From Reth**

```
gas: 0x55493 (349331)
gasUsed: 0x32d16 (208150)
```

**From [Etherscan](https://etherscan.io/vmtrace?txhash=0x03128677ee3a9623d20f3c677f423ccc592d126374bf32e331343dd1bdf38b61&type=parity#raw) and QuickNode**

```
gas: 0x4f227 (324135)
gasUsed: 0x36622 (222754)
```

## Solution for `revm-inspectors`

1. Not modify `gas_limit` and `gas_used` in the trace root

    ```diff
    - gas_limit = context.env.tx.gas_limit;
    - trace.set_root_trace_gas_used(gas_used);
    - trace.gas_used = gas_used(context.spec_id(), gas.spent(), gas.refunded() as u64);
    ```

2. The modification in Step 1 will cause another problem

    The `gas` field for Geth Debug Trace root will also be reset (not the gas limitation for the entire transaction).

    therefore, can define `set_transaction_gas_limit()` and `with_transaction_gas_limit()` for Geth Debug,

    which is similar to current `set_transaction_gas_used()` and `with_transaction_gas_used()` for Parity.

3. Then, modify the Reth Part: 

    `crates/rpc/rpc/src/trace.rs` and `crates/rpc/rpc/src/debug.rs` to completely fix the bug.

## Miscellaneous

- Actually, I love the current design, but the results are inconsistent with those of others.

- When I used `make pr` to test the Reth Part, the issue paradigmxyz/reth#9381 still exists for me.

    I should only skip tests for `lockfile` and test them seperately.

### DIFF
--- a/tests/it/geth.rs
+++ b/tests/it/geth.rs
@@ -306,7 +306,17 @@ fn test_geth_inspector_reset() {
     // first run inspector
     let (res, env) = inspect(&mut db, env.clone(), &mut insp).unwrap();
     assert!(res.result.is_success());
-    assert_eq!(insp.clone().with_transaction_gas_limit(env.tx.gas_limit).traces().nodes().first().unwrap().trace.gas_limit, 1000000);
+    assert_eq!(
+        insp.clone()
+            .with_transaction_gas_limit(env.tx.gas_limit)
+            .traces()
+            .nodes()
+            .first()
+            .unwrap()
+            .trace
+            .gas_limit,
+        1000000
+    );
 
     // reset the inspector
     insp.fuse();
@@ -315,5 +325,14 @@ fn test_geth_inspector_reset() {
     // second run inspector after reset
     let (res, env) = inspect(&mut db, env, &mut insp).unwrap();
     assert!(res.result.is_success());
-    assert_eq!(insp.with_transaction_gas_limit(env.tx.gas_limit).traces().nodes().first().unwrap().trace.gas_limit, 1000000);
+    assert_eq!(
+        insp.with_transaction_gas_limit(env.tx.gas_limit)
+            .traces()
+            .nodes()
+            .first()
+            .unwrap()
+            .trace
+            .gas_limit,
+        1000000
+    );
 }
--- a/tests/it/parity.rs
+++ b/tests/it/parity.rs
@@ -104,9 +104,8 @@ fn test_parity_selfdestruct(spec_id: SpecId) {
         assert_eq!(node.trace.selfdestruct_transferred_value, expected_value);
     }
 
-    let traces = insp
-        .into_parity_builder()
-        .into_localized_transaction_traces(TransactionInfo::default());
+    let traces =
+        insp.into_parity_builder().into_localized_transaction_traces(TransactionInfo::default());
 
     assert_eq!(traces.len(), 2);
     assert_eq!(
@@ -187,9 +186,8 @@ fn test_parity_constructor_selfdestruct() {
     assert!(res.result.is_success());
     print_traces(&insp);
 
-    let traces = insp
-        .into_parity_builder()
-        .into_localized_transaction_traces(TransactionInfo::default());
+    let traces =
+        insp.into_parity_builder().into_localized_transaction_traces(TransactionInfo::default());
 
     assert_eq!(traces.len(), 3);
     assert!(traces[1].trace.action.is_create());


src/tracing/mod.rs
```diff
@@ -173,6 +173,19 @@ impl TracingInspector {
         }
     }
 
+    /// Manually the gas limit of the debug root trace.
```
mattsse, line 176:
```suggestion
    /// Manually set the gas limit of the debug root trace.
```

mattsse (commented):
tysm for looking into this.

it seems like this is still a wip and I'll take a closer look once cleaned up.
but very supportive, your write up made a lot of sense

src/tracing/mod.rs
ZzPoLariszZ: Hi @mattsse, thanks so much for your quick comment. I have done some easy cleanup. Since I'm just getting started with Rust and Web3, please let me know somewhere can be improved or fixed. 😄

ZzPoLariszZ: Hi @mattsse, I think it is ready now.

src/tracing/mod.rs
```diff
@@ -173,6 +173,19 @@ impl TracingInspector {
         }
     }
 
+    /// Manually set the gas limit of the debug root trace.
```
mattsse, line 176:
```suggestion
    /// Manually set the gas limit of the root trace.
```

src/tracing/mod.rs
```diff
@@ -280,18 +300,6 @@ impl TracingInspector {
...
-        if self.trace_stack.is_empty() {
-            // this is the root call which should get the original gas limit of the transaction,
-            // because initialization costs are already subtracted from gas_limit
-            // For the root call this value should use the transaction's gas limit
-            // See <https://github.com/paradigmxyz/reth/issues/3678> and <https://github.com/ethereum/go-ethereum/pull/27029>
-            gas_limit = context.env.tx.gas_limit;
-
-            // we set the spec id here because we only need to do this once and this condition is
-            // hit exactly once
-            self.spec_id = Some(context.spec_id());
```
mattsse, line 292: this got lost, or do we no longer need this?

mattsse (changes requested):
ty!
last question

src/tracing/mod.rs
DaniPopes: Unused, we also have unused variables in the various constructors

src/tracing/mod.rs
ZzPoLariszZ: Hi @mattsse, the `debug` word here is to clarify this gas limit setting function only works for the Geth Debug trace root (to distinguish from the Parity trace root, which does not need to set this).

src/tracing/mod.rs
ZzPoLariszZ: I double check the code. This is `self.spec_id` for `TracingInspector` is used to set the gas limit for the Parity trace root only once. For Geth Debug trace, this will be set in the newly constructed function `set_transaction_gas_limit()` and `with_transaction_gas_limit()`. Therefore, we on longer need this.

mattsse (approved):
tysm!

doing a new release and we can finish the reth pr

Issue #65, opened 2024-03-09
Use borrowed Arena in GethTraceBuilder 


mattsse:
This should be a cow instead:

https://github.com/paradigmxyz/evm-inspectors/blob/846dec1db14ac42067e7a4bb6a973091cd7eb808/src/tracing/builder/geth.rs#L18-L20

together with:

https://github.com/paradigmxyz/evm-inspectors/pull/63

this would allows us to use an inspector across multiple transact calls, reusing allocated memory

nkysg:
Let me try it.

DaniPopes:
I believe this also applies to the other builders? @mattsse

PR #178, opened 2024-08-10
Use borrowed Arena in GethTraceBuilder

nkysg:
Closes https://github.com/paradigmxyz/revm-inspectors/issues/65


src/tracing/builder/geth.rs
```diff
@@ -11,20 +11,26 @@ use alloy_rpc_types::trace::geth::{
...
+    collections::{BTreeMap, HashMap, VecDeque},
+};
 
 /// A type for creating geth style traces
 #[derive(Clone, Debug)]
-pub struct GethTraceBuilder {
+pub struct GethTraceBuilder<'a> {
     /// Recorded trace nodes.
-    nodes: Vec<CallTraceNode>,
+    nodes: Cow<'a, Vec<CallTraceNode>>,
```
DaniPopes, line 23: should be [CallTraceNode]>

Cargo.toml
mattsse, line 1: could you submit this as a separate pr?

src/tracing/builder/geth.rs
```diff
@@ -11,20 +11,26 @@ use alloy_rpc_types::trace::geth::{
...
     /// How the traces were recorded
     _config: TracingInspectorConfig,
 }
 
-impl GethTraceBuilder {
+impl<'a> GethTraceBuilder<'a> {
     /// Returns a new instance of the builder
-    pub fn new(nodes: Vec<CallTraceNode>, _config: TracingInspectorConfig) -> Self {
+    pub fn new(
```
mattsse, line 30: I'd like to keep this function as is and add a new one `borrowed` instead

src/tracing/builder/geth.rs
```diff
@@ -11,20 +11,26 @@ use alloy_rpc_types::trace::geth::{
...
-use std::collections::{BTreeMap, HashMap, VecDeque};
+use std::{
+    borrow::Cow,
+    collections::{BTreeMap, HashMap, VecDeque},
+};
 
 /// A type for creating geth style traces
 #[derive(Clone, Debug)]
-pub struct GethTraceBuilder {
+pub struct GethTraceBuilder<'a> {
```
mattsse, line 21: This should then also implement `ToOwned`

src/tracing/builder/geth.rs
```diff
@@ -114,7 +120,7 @@ impl GethTraceBuilder {
     ///
     /// This expects the gas used and return value for the
     /// [ExecutionResult](revm::primitives::ExecutionResult) of the executed transaction.
-    pub fn geth_call_traces(&self, opts: CallConfig, gas_used: u64) -> CallFrame {
+    pub fn geth_call_traces(&self, opts: &CallConfig, gas_used: u64) -> CallFrame {
```
mattsse, line 123: this can stay `CallConfig` because this type is copy

mattsse (changes requested):
cool, some nits and suggestions

mattsse (approved):

PR #181, opened 2024-08-17
feat(tracing): add `unmatched` parameter to `CallLog`

topocount:
This allows for the rendering of unmatched logs as errors in traces
as discussed in foundry-rs/foundry#8506

### DIFF
--- a/src/tracing/types.rs
+++ b/src/tracing/types.rs
@@ -156,6 +156,17 @@ pub struct DecodedCallLog {
     pub params: Option<Vec<(String, String)>>,
 }
 
+/// portable representation of call Log Styling
+#[derive(Debug, Default, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum LogStyle {
+    /// The style when no special highlighting is needed
+    #[default]
+    Default,
+    /// The style when a log should be colored to denote an error
+    Error,
+}
+
 /// A log with optional decoded data.
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
@@ -164,8 +175,9 @@ pub struct CallLog {
     pub raw_log: LogData,
     /// Optional complementary decoded log data.
     pub decoded: DecodedCallLog,
-    /// A bool to render unmatched logs as errors from forge tests
-    pub unmatched: bool,
+    /// A style config that allows for setting styles or error
+    /// state externally
+    pub style: LogStyle,
 }
 
 impl From<Log> for CallLog {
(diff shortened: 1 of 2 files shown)


klkvr (changes requested):
I'd prefer this as configurable `style` key which foundry can set to `AnsiColor::Red`. Unmatched events logic is very foundry-specific, so don't think we should inroduce it here

src/tracing/types.rs
```diff
@@ -164,12 +175,19 @@ pub struct CallLog {
     pub raw_log: LogData,
     /// Optional complementary decoded log data.
     pub decoded: DecodedCallLog,
+    /// A style config that allows for setting styles or error
+    /// state externally
+    pub style: LogStyle,
```
klkvr, line 180: let's move this field to `decoded`

src/tracing/types.rs
```diff
@@ -156,6 +156,17 @@ pub struct DecodedCallLog {
     pub params: Option<Vec<(String, String)>>,
 }
 
+/// portable representation of call Log Styling
+#[derive(Debug, Default, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum LogStyle {
```
klkvr, line 162: this is fine for now, but I think we might want to give more flexibility here by keeping  some `anstyle` type here

we'd need to figure out serde for it, so can be done in a follow up

src/tracing/types.rs
```diff
@@ -156,6 +156,17 @@ pub struct DecodedCallLog {
     pub params: Option<Vec<(String, String)>>,
 }
 
+/// portable representation of call Log Styling
```
klkvr, line 159:
```suggestion
/// Portable representation of call log styling
```

klkvr (changes requested):

src/tracing/writer.rs
```diff
@@ -221,11 +220,13 @@ impl<W: Write> TraceWriter<W> {
...
     fn write_log(&mut self, log: &CallLog) -> io::Result<()> {
-        let log_style = self.log_style();
+        let log_name_style = self.log_name_style(log);
+        let log_param_style = self.log_param_style();
+
         self.write_branch()?;
 
         if let Some(name) = &log.decoded.name {
-            write!(self.writer, "emit {name}({log_style}")?;
+            write!(self.writer, "emit {log_name_style}{name}{log_name_style:#}({log_param_style}")?;
```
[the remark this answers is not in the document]
topocount: @klkvr are you okay with rendering ansi escape codes in terminals that don't support colors? Otherwise I think it makes sense to keep moving forward with the `log.style` impl?

zerosnacks:
Hi @topocount thanks for you PR, would be great to get this in - do you see an opportunity to implement the feedback?

topocount:
> Hi @topocount thanks for you PR, would be great to get this in - do you see an opportunity to implement the feedback?

@zerosnacks based on responses from @klkvr on foundry-rs/foundry#8506 i think this can be closed but im not completely sure

PR #184, opened 2024-08-23
fix: patch alloy and populate `log.position` for geth traces

klkvr:
`position` field has been added to `LogCallFrame`, so I've updated geth traces generation logic to populate it. Changed the loop to recursive function to allow us to use `node.ordering` for this

this currently patched to latest alloy


src/tracing/builder/geth.rs
```diff
@@ -119,58 +119,58 @@ impl GethTraceBuilder {
...
 
+        for item in node.ordering.iter().copied() {
             // selfdestructs are not recorded as individual call traces but are derived from
             // the call trace and are added as additional `CallFrame` objects
             // becoming the first child of the derived call
-            if let Some(selfdestruct) = trace.geth_selfdestruct_call_trace() {
-                call_frames.last_mut().expect("not empty").1.calls.push(selfdestruct);
+            if let Some(sefldestruct) = node.geth_selfdestruct_call_trace() {
+                call_frame.calls.push(sefldestruct);
             }
```
DaniPopes, line 151:
```suggestion
            if let Some(selfdestruct) = node.geth_selfdestruct_call_trace() {
                call_frame.calls.push(selfdestruct);
            }
```

PR #190, opened 2024-09-07
fix(tracing): don't overwrite selfdestruct_address

jsvisa:
fix https://github.com/paradigmxyz/reth/issues/10755

As selfdestruct reuse the call's struct, so we can't rewrite the call's `address`


mattsse (approved):
ty

PR #191, opened 2024-09-08
feat(tracing/builder): optimize the trace builder

jsvisa:
Some small optimizes:

- early return for StateDiff-only requests
- early check `if with_traces` instead of inside the for-loop
- for trace address use `Vec::with_capacity(n+1)` instead of `clone`

### DIFF
--- a/src/tracing/builder/parity.rs
+++ b/src/tracing/builder/parity.rs
@@ -252,7 +252,7 @@ impl ParityTraceBuilder {
 
             // Sort the traces only if a selfdestruct trace was encountered
             if sorting_selfdestruct {
-                traces.sort_by(|a, b| a.trace_address.cmp(&b.trace_address));
+                traces.sort_unstable_by(|a, b| a.trace_address.cmp(&b.trace_address));
             }
             traces
         });
--- a/src/tracing/builder/parity.rs
+++ b/src/tracing/builder/parity.rs
@@ -212,33 +212,33 @@ impl ParityTraceBuilder {
             return (None, None, None);
         }
 
-        let with_traces = trace_types.contains(&TraceType::Trace);
         let with_diff = trace_types.contains(&TraceType::StateDiff);
 
-        let vm_trace =
-            if trace_types.contains(&TraceType::VmTrace) { Some(self.vm_trace()) } else { None };
+        // early return for StateDiff-only case
+        if trace_types.len() == 1 && with_diff {
+            return (None, None, Some(StateDiff::default()));
+        }
 
-        let mut traces = Vec::with_capacity(if with_traces { self.nodes.len() } else { 0 });
-        // Boolean marker to track if sorting for selfdestruct is needed
-        let mut sorting_selfdestruct = false;
+        let vm_trace = trace_types.contains(&TraceType::VmTrace).then(|| self.vm_trace());
 
-        for node in self.iter_traceable_nodes() {
-            let trace_address = self.trace_address(node.idx);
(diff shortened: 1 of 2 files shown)


src/tracing/builder/parity.rs
```diff
@@ -249,15 +249,15 @@ impl ParityTraceBuilder {
...
             }
-        }
 
-        // Sort the traces only if a selfdestruct trace was encountered
-        if sorting_selfdestruct {
-            traces.sort_by(|a, b| a.trace_address.cmp(&b.trace_address));
-        }
+            // Sort the traces only if a selfdestruct trace was encountered
+            if sorting_selfdestruct {
+                traces.sort_by(|a, b| a.trace_address.cmp(&b.trace_address));
```
DaniPopes, line 255: this can be sort_by_key

src/tracing/types.rs
```diff
@@ -321,30 +317,22 @@ impl CallTraceNode {
...
-                typ: "SELFDESTRUCT".to_string(),
-                from: self.trace.address,
-                to: self.trace.selfdestruct_refund_target,
-                value: self.trace.selfdestruct_transferred_value,
-                ..Default::default()
-            })
-        } else {
-            None
-        }
+        self.is_selfdestruct().then_some(CallFrame {
```
DaniPopes, line 329: these then_some should be then(||

DaniPopes (changes requested):

src/tracing/builder/parity.rs
jsvisa: Seems like `sort_by_key` needs to consume the Key, ref https://users.rust-lang.org/t/why-does-slice-sort-by-key-uses-b-and-not-b-as-the-key/15781

this will need to copy the `trace_address`

```
error[E0507]: cannot move out of `t.trace_address` which is behind a shared reference
   --> src/tracing/builder/parity.rs:255:40
    |
255 |                 traces.sort_by_key(|t| t.trace_address);
    |                                        ^^^^^^^^^^^^^^^ move occurs because `t.trace_address` has type `Vec<usize>`, which does not implement the `Copy` trait
    |
help: consider cloning the value if the performance cost is acceptable
    |
255 |                 traces.sort_by_key(|t| t.trace_address.clone());
    |                                                       ++++++++

```

mattsse, line 255:
```suggestion
                traces.sort_unstable_by(|a, b| a.trace_address.cmp(&b.trace_address));
```

mattsse (changes requested):
nits

src/tracing/builder/parity.rs
```diff
@@ -212,33 +212,33 @@ impl ParityTraceBuilder {
...
 
-        let vm_trace =
-            if trace_types.contains(&TraceType::VmTrace) { Some(self.vm_trace()) } else { None };
+        // early return for StateDiff-only case
+        if trace_types.len() == 1 && with_diff {
+            return (None, None, Some(StateDiff::default()));
+        }
 
-        let mut traces = Vec::with_capacity(if with_traces { self.nodes.len() } else { 0 });
-        // Boolean marker to track if sorting for selfdestruct is needed
```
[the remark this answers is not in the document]
jsvisa: done

mattsse:
CI failing due to some unrelated version issue

src/tracing/builder/parity.rs
mattsse: yeah references also don't work for sort by key

mattsse (approved):

PR #196, opened 2024-09-12
feat: add helper for eth_simulateV1 to `TransferInspector`

klkvr:
ref https://github.com/ethereum/execution-apis/pull/484 https://github.com/paradigmxyz/reth/pull/10829

Also added hook for `eofcreate` and moved all handling away from `_env` hooks to ensure correct ordering.

We don't yet insert logs for selfdestructs, would need update on revm side to pass journaled state into the hook


mattsse (approved):
lgtm

mattsse:
clippy unrelated

PR #198, opened 2024-09-14
fix(tracing): align trace output with geth

pythonberg1997:
1. The error msg should not be converted to parity style in `callTracer`. They should only be converted in `flatCallTracer` which is not supported by reth now.
2. The `gasUsed` should be the `gasLimit` when there's a not reverted error.
3. The `call_frame.to` should be None if the call type is `Create/Create2` and it failed.

Related geth code: 
* https://github.com/ethereum/go-ethereum/blob/a9523b6428238a762e1a1e55e46ead47630c3a23/eth/tracers/native/call_flat.go#L226
* https://github.com/ethereum/go-ethereum/blob/a9523b6428238a762e1a1e55e46ead47630c3a23/core/vm/evm.go#L254
* https://github.com/ethereum/go-ethereum/blob/a9523b6428238a762e1a1e55e46ead47630c3a23/eth/tracers/native/call.go#L88

### DIFF
--- a/src/tracing/types.rs
+++ b/src/tracing/types.rs
@@ -140,6 +140,10 @@ impl CallTrace {
             InstructionResult::PrecompileError => {
                 if kind.is_parity() { "Built-in failed" } else { "precompiled failed" }.to_string()
             }
+            InstructionResult::InvalidFEOpcode => {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
ty, all of these look correct

PR #202, opened 2024-09-20
feat: Add creation code printing in traces

CodeSandwich:
This is needed for https://github.com/foundry-rs/foundry/issues/8906


mattsse (approved):
this seems fine, pending @klkvr

src/tracing/writer.rs
```diff
@@ -65,6 +67,13 @@ impl<W: Write> TraceWriter<W> {
         self
     }
 
+    /// Sets the starting indentation level.
```
DaniPopes, line 70: Update doc

src/tracing/writer.rs
```diff
@@ -65,6 +67,13 @@ impl<W: Write> TraceWriter<W> {
...
+    #[inline]
+    pub fn write_creation_codes(mut self, yes: bool) -> Self {
+        self.write_creation_codes = yes;
```
DaniPopes, line 73: I'd name it `code` instead of plural `codes`

DaniPopes (changes requested):

src/tracing/writer.rs
CodeSandwich: It follows the convention of `color_cheatcodes` because just like cheatcodes, creation codes can appear multiple times in a trace. If this reasoning doesn't convince you, I'll change it.

src/tracing/writer.rs
CodeSandwich: Fixed

tests/it/writer.rs
```diff
@@ -28,6 +28,16 @@ fn test_trace_printing() {
 "#]]
     );
 
+    let s = write_traces_with_creation_codes(&tracer);
+    let raw = r#"
+  [209257] → new <unknown>@0xBd770416a3345F91E4B34576cb804a576fa48EB1(<BYTECODE>)
+    └─ ← [Return] 1045 bytes of code
```
DaniPopes, line 34: should the deployed code be displayed too?

CodeSandwich: That's a good question. I need this feature because there's just no way to obtain the creation code, but the deployed code can be easily obtained after the transaction is finished unless there's some selfdestruction happening. I personally won't need the deployed code, but if you see value in printing it, I can add it too.

CodeSandwich: I think that it would actually be very neat even if I don't need it. I think that I should rename `write_creation_codes` to `with_bytecodes` and make that flag trigger printing both creation and deployed codes to avoid having too many knobs.

CodeSandwich: Pushed in the newest commit.

DaniPopes (approved):
LGTM, thanks!

klkvr (approved):

CodeSandwich:
It seems that I missed the 0.7.5 release by an hour :laughing: Do you have a plan to release 0.7.6 that would include changes from this PR?

mattsse:
doing rn

PR #203, opened 2024-09-22
chore: add `from_flat_call_config`

0xkrane:
For use and uniformity wrt https://github.com/paradigmxyz/reth/issues/10991

Simple tests included

I could do the PR without this but seems correct to maintain uniformity instead of modifying the `TracingInspector` in ad hoc ways

### DIFF
--- a/src/tracing/config.rs
+++ b/src/tracing/config.rs
@@ -400,4 +400,14 @@ mod tests {
         // not required for StateDiff
         assert!(!config.record_state_diff);
     }
+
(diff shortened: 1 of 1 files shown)


mattsse (approved):

mattsse:
we might need some helper function for the parity error setting, but lets ignore this for now in https://github.com/paradigmxyz/reth/issues/10991

0xkrane:
yeah was going to comment on the main PR, will take that up after

PR #218, opened 2024-10-05
chore(tracing): return detailed oog message

jsvisa:
Return a detailed OOG error, require https://github.com/bluealloy/revm/pull/1795 to be released first. ref https://github.com/ethereum/go-ethereum/pull/29354


mattsse (approved):
ty

PR #219, opened 2024-10-07
tracing/js: distinguish stack oob error

jsvisa:
distinguish the error of stack size out of bound or stake is null.


DaniPopes (approved):

mattsse (approved):

mattsse:
deny unrelated

PR #220, opened 2024-10-07
feat(js/tracing): invoke js's step inside step_end

### LINKED ISSUE #200: Fix gas cost for js tracer
reported: https://github.com/paradigmxyz/reth/issues/10959

we use incorrect gasCost value:

https://github.com/paradigmxyz/evm-inspectors/blob/0648673aab70f83722d5ad2832d023b87c0dcbbe/src/tracing/js/mod.rs#L401-L401

this should be gas cost of the opcode:

https://github.com/ethereum/go-ethereum/blob/0dd173a727dd2d2409b8e401b22e85d20c25b71f/core/vm/interpreter.go#L237-L237

this becomes a bit tricky for dynamic gas, because we don't know how much gas the opcode will cost if we call the tracer in the `step` fn, only if we track and call it on step_end.

see also:

https://github.com/paradigmxyz/evm-inspectors/blob/0648673aab70f83722d5ad2832d023b87c0dcbbe/src/opcode.rs#L78-L78

so we'd need to move this call to step_end, but this messes with memory because the tracer should be called with mem before opcode I believe...

I don't think there's an easier way to get this than doing

step: gas_remaining
step_end: gas_remaining - spent
@rakita

jsvisa:
fix https://github.com/paradigmxyz/revm-inspectors/issues/200 with the following adjusts:

1. invoke `step_fn` from `step` into `step_end`;
2. copy stake and memory at the ending of `step`;

waiting for https://github.com/paradigmxyz/revm-inspectors/pull/222 and https://github.com/bluealloy/revm/pull/1820 to be merged first.

~but failed with below issues: stake/memory related JS functions maybe failed.~

~Because we're invoking `step_fn.call` in step_end, seems the stack and memory are messed up, so if we try to call those functions inside JS's `step`(eg: `log.stack.peek(0)`) the JS binding will be failed, maybe it's caused by the `JsInspector::Context`, but as `boe_engine::Context` doesn't implemated Copy, we can't store the `step's Context` as the environment to used in the `step_end`~

### DIFF
--- a/src/tracing/js/bindings.rs
+++ b/src/tracing/js/bindings.rs
@@ -38,6 +38,8 @@ macro_rules! js_value_getter {
     };
 }
 
+pub(crate) use js_value_getter;
+
 /// A macro that creates a native function that returns a captured JsValue
 macro_rules! js_value_capture_getter {
     ($value:ident, $ctx:ident) => {
--- a/src/tracing/js/bindings.rs
+++ b/src/tracing/js/bindings.rs
@@ -237,9 +237,15 @@ impl StepLog {
 pub(crate) struct MemoryRef(GuardedNullableGc<SharedMemory>);
 
 impl MemoryRef {
-    /// Creates a new stack reference
+    /// Creates a new memory reference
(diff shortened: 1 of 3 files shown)


DaniPopes:
stale

PR #223, opened 2024-10-15
feat(tracing): js-tracer add coinbase into context

jsvisa:
add `coinbase` into js-tracer's context, backport https://github.com/ethereum/go-ethereum/pull/30231

### DIFF
--- a/src/tracing/js/bindings.rs
+++ b/src/tracing/js/bindings.rs
@@ -671,6 +671,8 @@ pub(crate) struct JsEvmContext {
     pub(crate) value: U256,
     /// Number, block number
     pub(crate) block: u64,
+    /// Address, miner of the block
+    pub(crate) coinbase: Address,
     pub(crate) output: Bytes,
     /// Number, block timestamp
     pub(crate) time: String,
@@ -692,6 +694,7 @@ impl JsEvmContext {
             intrinsic_gas,
             value,
             block,
+            coinbase,
             output,
             time,
             transaction_ctx,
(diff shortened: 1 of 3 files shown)


src/tracing/mod.rs
```diff
@@ -705,6 +705,10 @@ pub struct TransactionContext {
     ///
     /// `None` if this is a call.
     pub tx_hash: Option<B256>,
+    /// Address of the miner of the block.
+    ///
+    /// `None` if this is a call.
+    pub coinbase: Option<Address>,
```
mattsse, line 711: this we can't do because this should only contain external context

src/tracing/js/bindings.rs
```diff
@@ -727,6 +727,9 @@ impl JsEvmContext {
         if let Some(tx_hash) = transaction_ctx.tx_hash {
             obj.set(js_string!("txHash"), to_byte_array(tx_hash.0, ctx)?, false, ctx)?;
         }
+        if let Some(coinbase) = transaction_ctx.coinbase {
+            obj.set(js_string!("coinbase"), address_to_byte_array(coinbase, ctx)?, false, ctx)?;
```
mattsse, line 731: we need to get this from the `blockenv` instead:

https://github.com/paradigmxyz/revm-inspectors/blob/d55d3c156081980c803abd1c7062829aa9db8db9/src/tracing/js/mod.rs#L263-L263

via `env.blockenv.coinbase`

mattsse (changes requested):

src/tracing/js/bindings.rs
jsvisa: cool, we have `env.block`, so no need to adjust reth's codebase.

mattsse (approved):
ty!

PR #225, opened 2024-10-16
feat(tests): make the test code more clear and reuseable

jsvisa:
use TestEnv to reduce the duplication in test code.


mattsse (approved):
cool

PR #226, opened 2024-10-17
fix(tracing/js): toHex returns with 0x as prefix

jsvisa:
1. go-ethereum/erigon's `toHex` returns with `0x` prefix(ref https://github.com/ethereum/go-ethereum/blob/fad7e74a1b2995dd79002c184e532eff31382579/eth/tracers/js/goja.go#L467), we should keep the same;
2. add more unit tests for the builtin functions.


mattsse (approved):
ty

PR #231, opened 2024-10-17
chore(tracing/js): add more unit tests

jsvisa:
copy tests from https://github.com/ethereum/go-ethereum/blob/fad7e74a1b2995dd79002c184e532eff31382579/eth/tracers/js/tracer_test.go#L90


DaniPopes (approved):

jsvisa:
need https://github.com/paradigmxyz/revm-inspectors/pull/232 merged first, then the CIs will be succeed(I guess).

PR #235, opened 2024-10-21
feat: StackSnapshotType (All).

piyanggoon:
Add "StackSnapshotType::All" for easier reading of trace steps, as shown in the picture.

CleanShot 2567-10-22 at 03 05 05@2x


mattsse (approved):
reasonable to me.

PR #237, opened 2024-11-05
feat(trace/parity): add  trace creation method

jsvisa:
waiting for https://github.com/alloy-rs/alloy/pull/1621

### DIFF
--- a/src/tracing/types.rs
+++ b/src/tracing/types.rs
@@ -523,10 +523,9 @@ impl CallKind {
     }
 }
 
-/// Returns the creation method of the create callkind.
-impl Into<CreationMethod> for CallKind {
-    fn into(self) -> CreationMethod {
-        match self {
+impl From<CallKind> for CreationMethod {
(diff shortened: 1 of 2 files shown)


src/tracing/types.rs
```diff
@@ -522,6 +523,18 @@ impl CallKind {
     }
 }
 
+/// Returns the creation method of the create callkind.
+impl Into<CreationMethod> for CallKind {
+    fn into(self) -> CreationMethod {
```
mattsse, line 528: let's flip this to a from impl

src/tracing/types.rs
```diff
@@ -397,6 +397,7 @@ impl CallTraceNode {
                     value: self.trace.value,
                     gas: self.trace.gas_limit,
                     init: self.trace.data.clone(),
+                    creation_method: self.kind().into(),
```
mattsse, line 400: I see, this is only reachable for create

mattsse (approved):
lgtm, pending alloy bump today

Issue #201, opened 2024-09-18
Add support for flat call tracer in mux tracer


mattsse:
we now have flatcall tracer as builtin

https://github.com/paradigmxyz/revm-inspectors/blob/main/src/tracing/mux.rs#L233-L234

## TODO
* add support for this

this might perhaps be easier after #62

PR #240, opened 2024-11-23
tracing: implement FlatCallTracer

jsvisa:
closes https://github.com/paradigmxyz/revm-inspectors/issues/201

BTW, this will also change the public method `try_into_mux_frame`, so need to adjust in reth's side.


src/tracing/mux.rs
```diff
@@ -116,6 +126,17 @@ impl MuxInspector {
                         continue;
                     }
                 }
+                TraceConfig::FlatCall(_flatcall_config) => {
+                    if let Some(inspector) = &self.tracing {
+                        inspector
+                            .clone()
```
jsvisa, line 132: I'm curious if we can avoid clone here.

mattsse: we can't really because we need to consume in the parity tracer

mattsse (approved):
lgtm

PR #242, opened 2024-12-03
chore: remove bad todo

mattsse:
replaces todo! with a revert outcome return value


DaniPopes (approved):

mattsse:
bumping msrv seperately

PR #244, opened 2024-12-04
fix(js): incorrect caller and contract address extracting

jsvisa:
fixes https://github.com/paradigmxyz/reth/issues/13089

btw, this PR also fixes the incorrect caller address in delegatecall.


mattsse (approved):
awesome, tysm!

PR #246, opened 2024-12-20
chore: bump revm dec 2024

rakita:
Revm Framework integration


gakonst:
As discussed would like to hold on merging until we've iterated on the revm api a bit as this is used in a bunch of places and want to be thoughtful about breakages

rakita:
> As discussed would like to hold on merging until we've iterated on the revm api a bit as this is used in a bunch of places and want to be thoughtful about breakages

Doing exactly that, and I have already accounted interfaces and usage patterns when refactoring Revm. Changes that simplify integration even more, are here: https://github.com/bluealloy/revm/pull/1934

Present context allows us a lot simpler usage and it is easier to navigate.

Would assume this will not be merged until we have PR in Reth.

rakita:
@mattsse this is now integrated.

We should wait for Reth integration before merging.

Cargo.toml
```diff
@@ -34,7 +34,9 @@ alloy-rpc-types-eth = "0.11"
 alloy-rpc-types-trace = "0.11"
 alloy-sol-types = "0.8"
 alloy-primitives = { version = "0.8", features = ["map"] }
-revm = { version = "19.0.0", default-features = false, features = ["std"] }
+revm = { git = "https://github.com/bluealloy/revm.git", rev = "c12a95e8", default-features = false, features = [
```
klkvr, line 37:
```suggestion
revm = { git = "https://github.com/bluealloy/revm.git", rev = "b67ca3e", default-features = false, features = [
```

Cargo.toml
```diff
@@ -50,10 +52,20 @@ boa_gc = { version = "0.20", optional = true }
 
 [dev-dependencies]
 snapbox = { version = "0.6", features = ["term-svg"] }
+revm-database = { git = "https://github.com/bluealloy/revm.git", rev = "c12a95e8", default-features = false, features = [
```
klkvr, line 55:
```suggestion
revm-database = { git = "https://github.com/bluealloy/revm.git", rev = "b67ca3e", default-features = false, features = [
```

src/access_list.rs
```diff
@@ -65,67 +71,70 @@ impl AccessListInspector {
...
+    ) {
+        let from = context.tx().caller();
+        let to = if let TxKind::Call(to) = context.tx().kind() {
             to
         } else {
             // We need to exclude the created address if this is a CREATE frame.
             //
             // This assumes that caller has already been loaded but nonce was not increased yet.
-            let nonce = context.journaled_state.account(from).info.nonce;
+            let nonce = context.journal_ref().evm_state().get(&from).unwrap().info.nonce;
```
mattsse, line 85: nit, can we drop the _ref() suffix here? should just be journal and journal_mut

src/access_list.rs
```diff
@@ -65,67 +71,70 @@ impl AccessListInspector {
...
+        let precompiles = context.journal_ref().precompile_addresses().clone();
         self.excluded = [from, to].into_iter().chain(precompiles).collect();
     }
 }
 
-impl<DB> Inspector<DB> for AccessListInspector
+impl<CTX> Inspector<CTX, EthInterpreter> for AccessListInspector
 where
-    DB: Database,
+    CTX: ContextTrait<Journal: JournalExt>,
```
mattsse, line 95: naming a trait Trait always feels a bit weird to me

src/tracing/builder/geth.rs
```diff
@@ -216,7 +219,7 @@ impl<'a> GethTraceBuilder<'a> {
     /// * `db` - The database to fetch state pre-transaction execution.
     pub fn geth_prestate_traces<DB: DatabaseRef>(
         &self,
-        ResultAndState { state, .. }: &ResultAndState,
+        ResultAndState { state, .. }: &ResultAndState<impl HaltReasonTrait>,
```
mattsse, line 222: can we make this a named generic, don't like mixing this

src/tracing/builder/parity.rs
```diff
@@ -170,7 +173,7 @@ impl ParityTraceBuilder {
     /// with the [TracingInspector](crate::tracing::TracingInspector).
     pub fn into_trace_results_with_state<DB: DatabaseRef>(
         self,
-        res: &ResultAndState,
+        res: &ResultAndState<impl HaltReasonTrait>,
```
mattsse, line 176: same, can we name this just HaltReason or HaltReasonT

src/tracing/js/bindings.rs
```diff
@@ -945,33 +945,51 @@ struct EvmDbRefInner {
...
 }
 
 /// A wrapper Database for the JS context.
 pub(crate) struct JsDb<DB: DatabaseRef>(DB);
 
+#[derive(Clone, Debug)]
+pub(crate) struct StringError(pub String);
+
+impl core::error::Error for StringError {}
+impl DBErrorMarker for StringError {}
```
mattsse, line 958: why is this necessary?

src/tracing/js/bindings.rs
```diff
@@ -945,33 +945,51 @@ struct EvmDbRefInner {
...
     _state_guard: GcGuard<'a, EvmState>,
-    _db_guard: GcGuard<'b, Box<dyn DatabaseRef<Error = String> + 'static>>,
+    _db_guard: GcGuard<'b, Box<dyn DatabaseRef<Error = StringError> + 'static>>,
 }
 
 /// A wrapper Database for the JS context.
 pub(crate) struct JsDb<DB: DatabaseRef>(DB);
 
+#[derive(Clone, Debug)]
+pub(crate) struct StringError(pub String);
```
mattsse, line 955: needs docs, why this is necessary

mattsse, line 955: can we name this JsDbError

src/tracing/js/mod.rs
```diff
@@ -212,26 +218,29 @@ impl JsInspector {
...
 
     /// Calls the result function and returns the result.
-    pub fn result<DB>(
+    pub fn result<TX, DB>(
         &mut self,
-        res: ResultAndState,
-        env: &Env,
+        res: ResultAndState<impl HaltReasonTrait>,
+        tx: &TX,
+        block: &impl Block,
```
mattsse, line 239: can we make these consistent across all functions, don't like to mix this

src/access_list.rs
rakita: In most cases a mutable journal is needed, there was few cases were ref is needed so this fn was added

src/access_list.rs
rakita: Will rename all of them to `ContextT`.

src/access_list.rs
mattsse: I see, then I think it should remain like that for legacy reasons

mattsse (approved):

klkvr (approved):

PR #250, opened 2024-12-30
feat: support no_std

### LINKED ISSUE #248: Support no-std
we should be able to use no-std

## TODO
* add std feature
* use extern crate alloc
* update `use std` deps
* use alloy_primitives `features = ["map"]` type instead of std::collections

stevencartavia:
should close #248


Cargo.toml
```diff
@@ -30,19 +30,19 @@ clippy.lint_groups_priority = "allow"
...
-colorchoice = "1.0"
-thiserror = "2.0"
+anstyle = { version = "1.0", optional = true }
+colorchoice = { version = "1.0", optional = true }
+thiserror = { version = "2.0", default-features = false }
 
 # serde
 serde = { version = "1", optional = true, features = ["derive"] }
-serde_json = "1.0"
+serde_json = { version = "1.0", optional = true }
```
mattsse, line 45: this can use the `alloc` feature

Cargo.toml
```diff
@@ -52,5 +52,8 @@ boa_gc = { version = "0.20", optional = true }
 snapbox = { version = "0.6", features = ["term-svg"] }
 
 [features]
+default = ["std"]
+std = ["anstyle", "colorchoice", "serde_json"]
+no_std = []
```
mattsse, line 57: here we should also propagate std to everything

src/lib.rs
```diff
@@ -15,6 +15,11 @@
 #![deny(unused_must_use, rust_2018_idioms)]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
 
+extern crate alloc;
+
+#[cfg(feature = "std")]
+pub use colorchoice::ColorChoice;
```
mattsse, line 21: move below mods again

src/tracing/mux.rs
```diff
@@ -322,6 +323,10 @@ pub enum Error {
     #[error("expected config is missing for tracer '{0:?}'")]
     MissingConfig(GethDebugBuiltInTracerType),
     /// Error when deserializing the config
+    #[cfg(feature = "std")]
     #[error("error deserializing config: {0}")]
```
mattsse, line 327: I believe this type works in no-std as long as the serde-json alloc feature is enabled

src/tracing/writer.rs
```diff
@@ -5,13 +5,14 @@ use super::{
...
+#[cfg(feature = "std")]
 use anstyle::{AnsiColor, Color, Style};
+#[cfg(feature = "std")]
 use colorchoice::ColorChoice;
-use std::{
-    collections::HashMap,
-    io::{self, Write},
-};
+#[cfg(feature = "std")]
+use std::io::{self, Write};
```
mattsse, line 15: here we should rather feature gate the code that depends on this I assume

mattsse (changes requested):
cool,
the hardest part is already done I think, now we just need to deal with some features

Cargo.toml
```diff
@@ -33,16 +33,16 @@ clippy.lint_groups_priority = "allow"
 alloy-rpc-types-eth = "0.9"
 alloy-rpc-types-trace = "0.9"
 alloy-sol-types = "0.8"
-alloy-primitives = { version = "0.8", features = ["map"] }
+alloy-primitives = { version = "0.8", features = ["map-hashbrown"] }
```
mattsse, line 36: map should be sufficient here

Cargo.toml
```diff
@@ -52,5 +52,8 @@ boa_gc = { version = "0.20", optional = true }
 snapbox = { version = "0.6", features = ["term-svg"] }
 
 [features]
+default = ["std"]
+std = ["alloy-primitives/std", "anstyle/std", "serde/std", "serde_json/std", "revm/std", "thiserror/std"]
+no_std = []
```
mattsse, line 57: this isn't required because this is the same as no-default-features
```suggestion
```

src/lib.rs
```diff
@@ -4,7 +4,7 @@
 //!
 //! - `js-tracer`: Enables a JavaScript tracer implementation. This pulls in extra dependencies
 //!   (such as `boa`, `tokio` and `serde_json`).
-
+#![cfg_attr(not(feature = "std"), no_std)]
```
mattsse, line 7: move to L17

mattsse (changes requested):
few more nits

src/tracing/js/bindings.rs
DaniPopes, line 1: Please leave JS unchanged, it can never support no_std

DaniPopes (changes requested):

mattsse (approved):

PR #252, opened 2025-01-13
chore(tracer): no whitespace at the end of a line

jsvisa:
As some code editer will auto trim the ending whitespace, like below:

image

So when we're editing some testcases(eg: in the foundry project), it's annoyed for me to edit the test file, so I propose remove the ending backspace for the trace writer.


mattsse (approved):
lgtm

pending @DaniPopes

DaniPopes (approved):
fair

PR #255, opened 2025-01-20
feat: add edge coverage tracking inspired by AFL/Lucid

0xalpharush:
For use in Foundry coverage guided fuzzing. I think the other functionality such as [comparing two edge maps](https://github.com/h0mbre/Lucid/blob/3026e7323c52b30b3cf12563954ac1eaa9c6981e/src/coverage.rs#L72) can be added later or in Foundry directly

### DIFF
--- a/src/edge_cov.rs
+++ b/src/edge_cov.rs
@@ -1,0 +1,107 @@
+use alloy_primitives::{Address, U256};
+use revm::{
+    interpreter::{opcode::OpCode, Interpreter},
+    Database, EvmContext, Inspector,
+};
+
+use std::hash::{Hash, Hasher};
+
+/// A hit count that never goes to zero e.g. the back edge of a loop that iterates 256 times will be
+/// one instead of zero.
+// see https://github.com/AFLplusplus/AFLplusplus/blob/5777ceaf23f48ae4ceae60e4f3a79263802633c6/instrumentation/afl-llvm-pass.so.cc#L810-L829
+#[derive(Clone, Copy, Debug, Default)]
+pub struct NeverZeroHitCount(pub u8);
+
+impl std::ops::Add for NeverZeroHitCount {
+    type Output = Self;
+
+    fn add(self, rhs: Self) -> Self {
+        Self(self.0.checked_add(rhs.0).unwrap_or(1))
+    }
+}
+
+impl std::ops::AddAssign for NeverZeroHitCount {
+    fn add_assign(&mut self, rhs: Self) {
+        *self = *self + rhs;
+    }
+}
+
+const MAX_EDGE_COUNT: usize = 65536;
+
+/// An Inspector that tracks edge coverage.
+#[derive(Clone, Debug, Default)]
+pub struct EdgeCovInspector {
+    /// Map of total gas used per opcode.
+    pub hitcount: Vec<NeverZeroHitCount>,
+}
+
+impl EdgeCovInspector {
+    /// Create a new EdgeCovInspector.
+    pub fn new() -> Self {
+        Self { hitcount: vec![NeverZeroHitCount(0); MAX_EDGE_COUNT] }
+    }
+
+    /// Reset the hitcount to zero.
+    pub fn reset(&mut self) {
+        self.hitcount.fill(NeverZeroHitCount(0));
+    }
+}
+
+fn edge_hash(address: Address, pc: usize, jump_dest: U256) -> u32 {
+    // TODO faster hash https://h0mbre.github.io/Lucid_Snapshots_Coverage/
+    let mut hasher = std::collections::hash_map::DefaultHasher::new();
+    address.hash(&mut hasher);
+    pc.hash(&mut hasher);
+    jump_dest.hash(&mut hasher);
+    (hasher.finish() as usize % MAX_EDGE_COUNT) as u32
+}
+
+impl<DB> Inspector<DB> for EdgeCovInspector
+where
+    DB: Database,
+{
+    fn step(&mut self, interp: &mut Interpreter, _context: &mut EvmContext<DB>) {
+        let address = interp.contract.target_address; // TODO track context for delegatecall?
+        let current_pc = interp.program_counter();
+        let opcode_value = interp.current_opcode();
+        if let Some(op) = OpCode::new(opcode_value) {
+            match op {
+                OpCode::JUMP => {
+                    // unconditional jump
+                    if let Ok(jump_dest) = interp.stack().peek(0) {
+                        let edge_id = edge_hash(address, current_pc, jump_dest);
+                        self.hitcount[edge_id as usize] += NeverZeroHitCount(1);
+                    }
+                }
+                OpCode::JUMPI => {
+                    if let Ok(stack_value) = interp.stack().peek(0) {
+                        let jump_dest = if stack_value == U256::from(1) {
+                            // branch taken
+                            interp.stack().peek(1).unwrap()
+                        } else {
+                            // fall through
+                            U256::from(current_pc + 1)
+                        };
+                        let edge_id = edge_hash(address, current_pc, jump_dest);
+                        self.hitcount[edge_id as usize] += NeverZeroHitCount(1);
+                    }
+                }
+                _ => {
+                    // no-op
+                }
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_nonzero_hitcount() {
+        let mut hitcount = NeverZeroHitCount(255);
+        hitcount += NeverZeroHitCount(1);
+        assert_eq!(hitcount.0, 1);
+    }
+}
--- a/src/edge_cov.rs
+++ b/src/edge_cov.rs
@@ -1,42 +1,23 @@
 use alloy_primitives::{Address, U256};
+use core::hash::{Hash, Hasher};
 use revm::{
     interpreter::{opcode::OpCode, Interpreter},
     Database, EvmContext, Inspector,
 };
 
-use std::hash::{Hash, Hasher};
-
-/// A hit count that never goes to zero e.g. the back edge of a loop that iterates 256 times will be
-/// one instead of zero.
-// see https://github.com/AFLplusplus/AFLplusplus/blob/5777ceaf23f48ae4ceae60e4f3a79263802633c6/instrumentation/afl-llvm-pass.so.cc#L810-L829
-#[derive(Clone, Copy, Debug, Default)]
-pub struct NeverZeroHitCount(pub u8);
-
-impl std::ops::Add for NeverZeroHitCount {
-    type Output = Self;
-
-    fn add(self, rhs: Self) -> Self {
-        Self(self.0.checked_add(rhs.0).unwrap_or(1))
-    }
-}
-
-impl std::ops::AddAssign for NeverZeroHitCount {
-    fn add_assign(&mut self, rhs: Self) {
-        *self = *self + rhs;
-    }
-}
-
+// This is the maximum number of edges that can be tracked. There is a tradeoff between performance
+// and precision (less collisions).
 const MAX_EDGE_COUNT: usize = 65536;
 
-/// An Inspector that tracks edge coverage.
+/// An `Inspector` that tracks [edge coverage](https://clang.llvm.org/docs/SanitizerCoverage.html#edge-coverage).
 #[derive(Clone, Debug, Default)]
 pub struct EdgeCovInspector {
-    /// Map of total gas used per opcode.
-    pub hitcount: Vec<NeverZeroHitCount>,
+    /// Map of hitcounts that can be diffed against to determine if new coverage was reached.
+    hitcount: Vec<NeverZeroHitCount>,
 }
 
 impl EdgeCovInspector {
-    /// Create a new EdgeCovInspector.
+    /// Create a new `EdgeCovInspector` with `MAX_EDGE_COUNT` size.
     pub fn new() -> Self {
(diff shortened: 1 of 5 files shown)


src/edge_cov.rs
```diff
@@ -0,0 +1,107 @@
+use alloy_primitives::{Address, U256};
+use revm::{
+    interpreter::{opcode::OpCode, Interpreter},
+    Database, EvmContext, Inspector,
+};
+
+use std::hash::{Hash, Hasher};
```
mattsse, line 7:
```suggestion
use core::hash::{Hash, Hasher};
```

src/edge_cov.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn add(self, rhs: Self) -> Self {
+        Self(self.0.checked_add(rhs.0).unwrap_or(1))
+    }
+}
+
+impl std::ops::AddAssign for NeverZeroHitCount {
+    fn add_assign(&mut self, rhs: Self) {
+        *self = *self + rhs;
+    }
+}
```
mattsse, line 27: can we move this below the inspector type and impl

src/edge_cov.rs
```diff
@@ -0,0 +1,107 @@
...
+
+const MAX_EDGE_COUNT: usize = 65536;
+
+/// An Inspector that tracks edge coverage.
+#[derive(Clone, Debug, Default)]
+pub struct EdgeCovInspector {
+    /// Map of total gas used per opcode.
```
mattsse, line 34: copy paste leftover?

src/edge_cov.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    /// Reset the hitcount to zero.
+    pub fn reset(&mut self) {
+        self.hitcount.fill(NeverZeroHitCount(0));
+    }
+}
+
+fn edge_hash(address: Address, pc: usize, jump_dest: U256) -> u32 {
+    // TODO faster hash https://h0mbre.github.io/Lucid_Snapshots_Coverage/
+    let mut hasher = std::collections::hash_map::DefaultHasher::new();
```
mattsse, line 52: we could also use alloy_primitives::map::DefaultHasher here

src/edge_cov.rs
```diff
@@ -0,0 +1,107 @@
...
+
+impl std::ops::AddAssign for NeverZeroHitCount {
+    fn add_assign(&mut self, rhs: Self) {
+        *self = *self + rhs;
+    }
+}
+
+const MAX_EDGE_COUNT: usize = 65536;
+
+/// An Inspector that tracks edge coverage.
```
mattsse, line 31: not immediately obvious what this means.
can we add a few more docs here that describe what this records?

src/edge_cov.rs
```diff
@@ -0,0 +1,107 @@
...
+    /// Create a new EdgeCovInspector.
+    pub fn new() -> Self {
+        Self { hitcount: vec![NeverZeroHitCount(0); MAX_EDGE_COUNT] }
+    }
+
+    /// Reset the hitcount to zero.
+    pub fn reset(&mut self) {
+        self.hitcount.fill(NeverZeroHitCount(0));
+    }
+}
```
mattsse, line 48: should there be a fn to consume the hitcount

src/edge_cov.rs
```diff
@@ -0,0 +1,107 @@
...
+
+const MAX_EDGE_COUNT: usize = 65536;
+
+/// An Inspector that tracks edge coverage.
+#[derive(Clone, Debug, Default)]
+pub struct EdgeCovInspector {
+    /// Map of total gas used per opcode.
+    pub hitcount: Vec<NeverZeroHitCount>,
```
mattsse, line 35: having this pub is a bit dangerous because this way a user can violate the non empty invariant

mattsse (changes requested):

src/edge_cov.rs
DaniPopes: should be DefaultHashBuilder::default().build_hasher()

mattsse: I think @DaniPopes is suggesting

`alloy_primitives::map::DefaultHashBuilder::default().build_hasher()`

the default hasher is chosen at compile time, depending on the features

mattsse (changes requested):

src/edge_cov.rs
0xalpharush: I tried this hasher and it doesn't produce the same edge hash for the same inputs. It seems there is a per hasher seed. Ig maybe there's some way to make this global and have the inspector maintain the state?

DaniPopes: ah right of course, you want to store the hash builder in the struct and use that to .build_hasher()

src/edge_cov.rs
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+    }
+}
+
+/// A hit count that never goes to zero e.g. the back edge of a loop that iterates 256 times will be
+/// one instead of zero.
+// see https://github.com/AFLplusplus/AFLplusplus/blob/5777ceaf23f48ae4ceae60e4f3a79263802633c6/instrumentation/afl-llvm-pass.so.cc#L810-L829
+#[derive(Clone, Copy, Debug, Default)]
+pub struct NeverZeroHitCount(pub u8);
```
DaniPopes, line 96: NonZeroU8?

src/edge_cov.rs
```diff
@@ -0,0 +1,122 @@
...
+
+impl<DB> Inspector<DB> for EdgeCovInspector
+where
+    DB: Database,
+{
+    fn step(&mut self, interp: &mut Interpreter, _context: &mut EvmContext<DB>) {
+        let address = interp.contract.target_address; // TODO track context for delegatecall?
+        let current_pc = interp.program_counter();
+        let opcode_value = interp.current_opcode();
+        if let Some(op) = OpCode::new(opcode_value) {
```
DaniPopes, line 62: should just match on opcode with opcode::JUMP (the u8 value directly)

src/edge_cov.rs
```diff
@@ -0,0 +1,122 @@
...
+                OpCode::JUMP => {
+                    // unconditional jump
+                    if let Ok(jump_dest) = interp.stack().peek(0) {
+                        let edge_id = edge_hash(address, current_pc, jump_dest);
+                        self.hitcount[edge_id as usize] += NeverZeroHitCount(1);
+                    }
+                }
+                OpCode::JUMPI => {
+                    if let Ok(stack_value) = interp.stack().peek(0) {
+                        let jump_dest = if stack_value == U256::from(1) {
```
DaniPopes, line 73: should be != 0

src/edge_cov.rs
```diff
@@ -0,0 +1,122 @@
...
+                            // branch taken
+                            interp.stack().peek(1).unwrap()
```
DaniPopes, line 75: should not unwrap

src/edge_cov.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    /// Reset the hitcount to zero.
+    pub fn reset(&mut self) {
+        self.hitcount.fill(NeverZeroHitCount(0));
+    }
+
+    /// Get the hitcount as a byte vector.
+    pub fn get_hitcount(&self) -> Vec<u8> {
+        self.hitcount.iter().map(|x| x.0).collect()
+    }
```
DaniPopes, line 33: should return a slice instead of copying

src/edge_cov.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/// A hit count that never goes to zero e.g. the back edge of a loop that iterates 256 times will be
+/// one instead of zero.
+// see https://github.com/AFLplusplus/AFLplusplus/blob/5777ceaf23f48ae4ceae60e4f3a79263802633c6/instrumentation/afl-llvm-pass.so.cc#L810-L829
+#[derive(Clone, Copy, Debug, Default)]
+pub struct NeverZeroHitCount(pub u8);
+
+impl std::ops::Add for NeverZeroHitCount {
```
DaniPopes, line 98: i think it should just expose .get() -> u8 and .increment(&mut self)

DaniPopes (changes requested):

src/edge_cov.rs
0xalpharush: We still need to be able to represent zero as not-yet-hit

src/edge_cov.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn step(&mut self, interp: &mut Interpreter, _context: &mut EvmContext<DB>) {
+        let address = interp.contract.target_address; // TODO track context for delegatecall?
+        let current_pc = interp.program_counter();
+
+        match interp.current_opcode() {
+            opcode::JUMP => {
+                // unconditional jump
+                if let Ok(jump_dest) = interp.stack().peek(0) {
+                    let edge_id = self.edge_hash(address, current_pc, jump_dest) as usize;
+                    self.hitcount[edge_id] = self.hitcount[edge_id].checked_add(1).unwrap_or(1);
```
0xalpharush, line 72: I went ahead and removed `NeverZeroHitCount` bc returning a slice require an unsafe block and sort of defeated any advantage a custom type would have. It's cleaner to just handle the wrapping case locally

src/edge_cov.rs
DaniPopes: right

src/edge_cov.rs
```diff
@@ -0,0 +1,94 @@
...
+                if let Ok(stack_value) = interp.stack().peek(0) {
+                    if let Ok(jump_dest) = if stack_value != U256::from(0) {
+                        // branch taken
+                        interp.stack().peek(1)
+                    } else {
+                        // fall through
+                        Ok(U256::from(current_pc + 1))
+                    } {
+                        let edge_id = self.edge_hash(address, current_pc, jump_dest) as usize;
+                        self.hitcount[edge_id] = self.hitcount[edge_id].checked_add(1).unwrap_or(1);
```
DaniPopes, line 85: we can factor out edge_hash + increment into a separate function

src/edge_cov.rs
```diff
@@ -0,0 +1,94 @@
...
+            opcode::JUMP => {
+                // unconditional jump
+                if let Ok(jump_dest) = interp.stack().peek(0) {
+                    let edge_id = self.edge_hash(address, current_pc, jump_dest) as usize;
+                    self.hitcount[edge_id] = self.hitcount[edge_id].checked_add(1).unwrap_or(1);
+                }
+            }
+            opcode::JUMPI => {
+                if let Ok(stack_value) = interp.stack().peek(0) {
+                    if let Ok(jump_dest) = if stack_value != U256::from(0) {
```
DaniPopes, line 77: please extract the inner if-else, this is not very readable

src/edge_cov.rs
```diff
@@ -0,0 +1,94 @@
...
+        Self { hitcount: vec![0; MAX_EDGE_COUNT], hash_builder: DefaultHashBuilder::default() }
+    }
+
+    /// Reset the hitcount to zero.
+    pub fn reset(&mut self) {
+        self.hitcount.fill(0);
+    }
+
+    /// Get the hitcount as a byte vector.
+    pub fn get_hitcount(&self) -> &[u8] {
```
DaniPopes, line 37: how can this be used outside of this crate? should there be a getter that uses edge_hash ?

0xalpharush: The fuzzer would maintain its own history for the edges and diff against the latest one (returned by `get_hitcount`) after execution. The actually index doesn't need to be known, just whether it's an "interesting change" that should be put in the corpus. https://github.com/h0mbre/Lucid/blob/3026e7323c52b30b3cf12563954ac1eaa9c6981e/src/coverage.rs#L58-L85

tests/it/edge_cov.rs
```diff
@@ -0,0 +1,108 @@
...
+            .into(),
+        ..Default::default()
+    };
+    let env = EnvWithHandlerCfg::new_with_cfg_env(cfg.clone(), BlockEnv::default(), tx_env.clone());
+    insp.reset();
+    let (res, _) = inspect(&mut db, env, &mut insp).unwrap();
+    assert!(res.result.is_success());
+
+    // There should be 13 non-zero counts and two edges that have been hit 255 times.
+    let mut counts = insp.get_hitcount().to_owned();
```
mattsse, line 103: there should be a fn to consume the inspector and return the hitcount vec

mattsse (changes requested):
last suggestion,

pending @DaniPopes

DaniPopes (approved):
lgtm, missing no_std compatibility

src/edge_cov.rs
```diff
@@ -33,11 +33,16 @@ impl EdgeCovInspector {
...
-    /// Get the hitcount as a byte vector.
+    /// Get an immutable reference to the hitcount.
     pub fn get_hitcount(&self) -> &[u8] {
         self.hitcount.as_slice()
     }
 
+    /// Consume the inspector and take ownership of the hitcount..
+    pub fn take_hitcount(&mut self) -> Vec<u8> {
+        std::mem::take(&mut self.hitcount)
+    }
```
mattsse: this is also problematic because this violates the non-empty constraint

```suggestion
    /// Consume the inspector and take ownership of the hitcount..
    pub fn into_hitcount(self) -> Vec<u8> {
        self.hitcount
    }
```

mattsse (changes requested):

src/edge_cov.rs
0xalpharush: I misunderstood your comment.

Btw it's fine if it's zero. We just don't want 255+1 to be represented as 0. The `edge_hash` function ensures that
https://github.com/paradigmxyz/revm-inspectors/pull/255#discussion_r1924132534

mattsse (approved):

PR #256, opened 2025-01-24
feat: simplify AccessListInspector API

klkvr:
Right now caller is required to provide `AccessListInspector` with the list of precompiles which is unnecessary and requires caller to peek into EVM internals. Besides making API more complex this also results in reth not excluding p256 OP precompile when preparing access list https://github.com/paradigmxyz/reth/blob/890507a98adef94b3970312ef796eb40f87206df/crates/rpc/rpc-eth-api/src/helpers/call.rs#L430-L431

This PR changes `AccessListInspector` to obtain `from`, `to` and `precompiles` from the top-level frame and not requiring caller to provide anything

### DIFF
--- a/src/access_list.rs
+++ b/src/access_list.rs
@@ -1,3 +1,4 @@
+use alloc::collections::BTreeSet;
 use alloy_primitives::{
     map::{HashMap, HashSet},
     Address, TxKind, B256,
@@ -7,7 +8,6 @@ use revm::{
     interpreter::{opcode, Interpreter},
     Database, EvmContext, Inspector,
 };
-use std::collections::BTreeSet;
 
 /// An [Inspector] that collects touched accounts and storage slots.
 ///
--- a/src/access_list.rs
+++ b/src/access_list.rs
@@ -1,13 +1,13 @@
-use alloc::collections::BTreeSet;
 use alloy_primitives::{
     map::{HashMap, HashSet},
-    Address, B256,
+    Address, TxKind, B256,
 };
 use alloy_rpc_types_eth::{AccessList, AccessListItem};
 use revm::{
     interpreter::{opcode, Interpreter},
     Database, EvmContext, Inspector,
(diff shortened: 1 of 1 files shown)


src/access_list.rs
```diff
@@ -59,12 +54,63 @@ impl AccessListInspector {
         });
         AccessList(items.collect())
     }
+
+    /// Collects addresses which should be excluded from the access list.
+    ///
+    /// Those include caller, callee and precompiles.
+    fn collect_excluded_addresses<DB: Database>(&mut self, context: &EvmContext<DB>) {
```
mattsse, line 61: can we add a note here that we must collect this on depth 0?

src/access_list.rs
```diff
@@ -59,12 +54,63 @@ impl AccessListInspector {
...
+        let from = context.env.tx.caller;
+        let to = if let TxKind::Call(to) = context.env.tx.transact_to {
+            to
+        } else {
+            let nonce = context.journaled_state.account(from).info.nonce;
+            from.create(nonce)
```
mattsse, line 67: let's add sanity note here that this will be a create call and we should exclude the created address

src/access_list.rs
```diff
@@ -24,14 +24,9 @@ impl AccessListInspector {
     /// Creates a new inspector instance
     ///
     /// The `access_list` is the provided access list from the call request
-    pub fn new(
-        access_list: AccessList,
-        from: Address,
-        to: Address,
-        precompiles: impl IntoIterator<Item = Address>,
-    ) -> Self {
+    pub fn new(access_list: AccessList) -> Self {
```
mattsse, line 27: could we now also add a From impl

mattsse (changes requested):
this makes a lot of sense, the overhead of checking depth 0 is negligible imo

mattsse (approved):

Issue #267, opened 2025-03-07
Use accurate refund counter


mattsse:
as described here

https://github.com/paradigmxyz/reth/issues/14783

we're incorrectly tracking refunds because revm's refund function returns the refund per call which can go negative.

but we want the total refund counter.

to fix this we need to change the type to an i64 because this can go negative:

https://github.com/paradigmxyz/revm-inspectors/blob/c847b12c470a84e677a7c93ce98fc0dea1ab0b3e/src/tracing/types.rs#L674-L675

and move this

https://github.com/paradigmxyz/revm-inspectors/blob/c847b12c470a84e677a7c93ce98fc0dea1ab0b3e/src/tracing/mod.rs#L445-L445

to the step_end function, here:

https://github.com/paradigmxyz/revm-inspectors/blob/c847b12c470a84e677a7c93ce98fc0dea1ab0b3e/src/tracing/mod.rs#L519-L519

this now only tracks the refund snapshot on call level, so what we need to do on `call_end` is to compute the total refund counter by looking at the previous calls

https://github.com/paradigmxyz/revm-inspectors/blob/c847b12c470a84e677a7c93ce98fc0dea1ab0b3e/src/tracing/mod.rs#L341-L341

and compute the real refund counters

paronikyanarmen:
Will work on this

PR #268, opened 2025-03-13
[DRAFT] Fix refund calculation

paronikyanarmen:
# Fix refund calculation

## Fixes
Fixes issue: #267

## Description
Should allow refund value in a call to be a negative value, since EVM supports negative values

### DIFF
--- a/src/tracing/types.rs
+++ b/src/tracing/types.rs
@@ -97,6 +97,8 @@ pub struct CallTrace {
     pub steps: Vec<CallTraceStep>,
     /// Optional complementary decoded call data.
     pub decoded: DecodedCallTrace,
+    /// The total gas refunded in the call.
+    pub gas_refunded: u64
 }
 
 impl CallTrace {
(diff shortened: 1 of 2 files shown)


src/tracing/types.rs
```diff
@@ -707,7 +709,7 @@ impl CallTraceStep {
             gas_cost: self.gas_cost,
             op: self.op.to_string(),
             pc: self.pc as u64,
-            refund_counter: (self.gas_refund_counter > 0).then_some(self.gas_refund_counter),
+            refund_counter: (self.gas_refund_counter > 0).then_some(self.gas_refund_counter as u64),
```
mattsse, line 712: I think we can do it like this but only if we update the refund counters accordingly, because we're now only tracking the refund on a per call level, so what we need to di is one post processing step that updates the counter when a call ended.


I believe we could do this when we initialize the first calltrace using the sum of all gas refunds up until this point.

mattsse (changes requested):
great, all of these look good so far, the more challenging part is tracking the total gas refunds per call which I think we need to do when we create a new call object 

but maybe there are some things we also need to consider on call_end @rakita ?

rakita:
Yeah, there should be a `global_refund_counter` that gets updated added the refund of subcall in `call_end`, `create_end`, `eofcreate_end`. 

And traced refund should be a `self.global_refund_counter+interp.control.gas().refunded()` it makes sense for this to be calculated on `step_end`

DaniPopes:
stale

PR #273, opened 2025-03-24
fix: populate selfdestructs in localized parity

mattsse:
closes https://github.com/paradigmxyz/reth/issues/15150

the bug here was that we didn't record all selfdestructs and replaced previous ones (higher address) if there was another selfdestruct.

this fixes this by recording selfdestructs in a stack

### DIFF
--- a/src/tracing/builder/parity.rs
+++ b/src/tracing/builder/parity.rs
@@ -427,10 +427,10 @@ where
             // find the next selfdestruct to yield
             if let Some((next_trace, _)) = self.iter.peek() {
                 // find the most recently recorded selfdestruct that has a lower address
-                if let Some(pos) =
-                    self.next_selfdestructs.iter().rev().position(|selfdestruct| {
-                        selfdestruct.trace_address < next_trace.trace_address
-                    })
+                if let Some(pos) = self
+                    .next_selfdestructs
+                    .iter()
+                    .rposition(|selfdestruct| selfdestruct.trace_address < next_trace.trace_address)
                 {
                     return Some(self.next_selfdestructs.remove(pos));
                 }
--- a/src/tracing/builder/parity.rs
+++ b/src/tracing/builder/parity.rs
@@ -224,53 +224,59 @@ impl ParityTraceBuilder {
 
         let vm_trace = trace_types.contains(&TraceType::VmTrace).then(|| self.vm_trace());
 
-        let traces = trace_types.contains(&TraceType::Trace).then(|| {
-            let mut traces = Vec::with_capacity(self.nodes.len());
-            // Boolean marker to track if sorting for selfdestruct is needed
-            let mut sorting_selfdestruct = false;
-
-            for node in self.iter_traceable_nodes() {
-                let trace_address = self.trace_address(node.idx);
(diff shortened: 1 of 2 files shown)


src/tracing/builder/parity.rs
```diff
@@ -224,53 +224,59 @@ impl ParityTraceBuilder {
...
         let diff = with_diff.then(StateDiff::default);
 
         (traces, vm_trace, diff)
     }
 
+    /// Returns all the ordered [`TransactionTrace`], including selfdestructs.
+    ///
+    /// Selfdestructs appear as individual [`TransactionTrace`] instance but selfdestructs are
+    /// tracked as metadata of the recorded nodes.
+    fn transaction_traces(&self) -> Vec<TransactionTrace> {
```
mattsse, line 238: I only extracted this so I can properly document this.

and originally I thought we'd need to use this but we can also solve this with the iter, so no functional changes here

src/tracing/builder/parity.rs
```diff
@@ -401,8 +407,12 @@ impl ParityTraceBuilder {
 
 /// An iterator for [TransactionTrace]s
 struct TransactionTraceIter<Iter: Iterator> {
+    /// The iterator over all traces
     iter: Peekable<Iter>,
-    next_selfdestruct: Option<TransactionTrace>,
+    /// The selfdestruct objects that are derived from the yielded traces.
+    ///
+    /// This is a stack because we need to yield them in the correct order.
+    next_selfdestructs: Vec<TransactionTrace>,
```
mattsse, line 415: this is the actual fix here

src/tracing/builder/parity.rs
```diff
@@ -413,13 +423,20 @@ where
...
+        if !self.next_selfdestructs.is_empty() {
+            // find the next selfdestruct to yield
+            if let Some((next_trace, _)) = self.iter.peek() {
+                // find the most recently recorded selfdestruct that has a lower address
+                if let Some(pos) =
+                    self.next_selfdestructs.iter().rev().position(|selfdestruct| {
+                        selfdestruct.trace_address < next_trace.trace_address
+                    })
+                {
+                    return Some(self.next_selfdestructs.remove(pos));
```
DaniPopes, line 435: `pos` is not an index into the vec but a count from the end

DaniPopes (changes requested):

tests/it/parity.rs
```diff
@@ -245,6 +245,84 @@ fn test_parity_call_selfdestruct() {
     );
 }
 
+// Minimal example of <https://github.com/paradigmxyz/reth/issues/15150, <0x0b0c51740c9fa9f6b9120410ccaac2eb51b81200a64b6fe0b886c762eb03f48b>
```
DaniPopes, line 248:
```suggestion
// Minimal example of <https://github.com/paradigmxyz/reth/issues/15150>, 0x0b0c51740c9fa9f6b9120410ccaac2eb51b81200a64b6fe0b886c762eb03f48b
```

src/tracing/builder/parity.rs
mattsse: right

DaniPopes (approved):

PR #277, opened 2025-04-04
Disable Revm default features

ufoscout:
`revm-inspector` imports `revm` with the default features enabled. This causes an issue where it's not possible to disable Revm's default features when this crate is imported. This PR resolves the problem by explicitly disabling those features, as they are not required by `revm-inspector`.


mattsse (approved):
good change, ty!

PR #280, opened 2025-04-09
chore: alloy 0.14

yash-atreya:
Bumps 

- [x] alloy-core 1.0
- [x] alloy 0.14
- [x] revm


src/tracing/js/mod.rs
```diff
@@ -628,11 +628,8 @@ pub enum JsInspectorError {
 /// Converts a JavaScript error into a [InstructionResult::Revert] [InterpreterResult].
 #[inline]
 fn js_error_to_revert(err: JsError) -> InterpreterResult {
-    InterpreterResult {
-        result: InstructionResult::Revert,
-        output: err.to_string().into(),
-        gas: Gas::new(0),
-    }
+    let output = err.to_string().as_bytes().to_vec();
```
mattsse, line 631: ah right we removed th from string impl

mattsse (approved):

PR #285, opened 2025-04-29
feat: added storageinspector

Rimeeeeee:
Towards [#15984](https://github.com/paradigmxyz/reth/issues/15984)

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -1,0 +1,66 @@
+use alloc::collections::BTreeSet;
+use alloy_primitives::{map::HashMap, Address, B256};
+use revm::{
+    bytecode::opcode,
+    context::ContextTr,
+    inspector::JournalExt,
+    interpreter::{
+        interpreter_types::{InputsTr, Jumps},
+        Interpreter,
+    },
+    Inspector,
+};
+
+/// An Inspector that tracks warm and cold storage slot accesses.
+#[derive(Debug, Default)]
+pub struct StorageInspector {
+    /// Tracks which storage slots have been accessed
+    accessed_slots: HashMap<Address, BTreeSet<B256>>,
+    /// Counter for cold SLOADs
+    cold_loads: u64,
+    /// Counter for warm SLOADs  
+    warm_loads: u64,
+}
+
+impl StorageInspector {
+    /// Creates a new storage inspector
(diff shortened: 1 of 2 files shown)


src/storage.rs
```diff
@@ -0,0 +1,66 @@
...
+
+/// An Inspector that tracks warm and cold storage slot accesses.
+#[derive(Debug, Default)]
+pub struct StorageInspector {
+    /// Tracks which storage slots have been accessed
+    accessed_slots: HashMap<Address, BTreeSet<B256>>,
+    /// Counter for cold SLOADs
+    cold_loads: u64,
+    /// Counter for warm SLOADs  
+    warm_loads: u64,
```
mattsse, line 22: cool, we can track this as part of the address, we can simply convert these two fields into a new struct and convert BTreeSet<B256> into a map

src/storage.rs
```diff
@@ -0,0 +1,66 @@
...
+                let address = interp.input.target_address();
+                let slot = B256::from(slot.to_be_bytes());
+
+                // Check if this slot was previously accessed
+                if let Some(slots) = self.accessed_slots.get(&address) {
+                    if slots.contains(&slot) {
+                        self.warm_loads += 1;
+                        return;
+                    }
+                }
```
mattsse, line 58: this should actually do it, nice

mattsse (changes requested):
great, only some structural nits

src/storage.rs
```diff
@@ -0,0 +1,77 @@
...
+    cold_loads: u64,
+    /// Number of times this slot was accessed when warm
+    warm_loads: u64,
+}
+
+/// An Inspector that tracks warm and cold storage slot accesses.
+#[derive(Debug, Default)]
+pub struct StorageInspector {
+    /// Tracks storage slots and their access statistics per address
+    accessed_slots: HashMap<Address, HashMap<B256, SlotStats>>,
```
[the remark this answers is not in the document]
Rimeeeeee: should i revert to the initial way of defining the cold_loads and warm_loads? if i need to modify it somehow lmk :)

mattsse (approved):
very nice ty,

just waiting to merge this in case we need some additonal features

PR #287, opened 2025-04-30
chore(access-list): Add function to access touched slots

cakevm:
Since there is no way to access the internal map, this PR introduces a `fn touched_slots`. To avoid conflicting with the semantics of `fn access_list`, I named it `touched_slots`. The reason for this is that using the map speeds up the process of checking whether certain slots have been touched.

### DIFF
--- a/src/access_list.rs
+++ b/src/access_list.rs
@@ -60,6 +60,12 @@ impl AccessListInspector {
         &self.access_list
     }
 
+    /// Consumes the inspector and returns the map of addresses and their corresponding touched
+    /// storage slots.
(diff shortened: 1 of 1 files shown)


src/access_list.rs
```diff
@@ -26,7 +26,7 @@ pub struct AccessListInspector {
     /// All addresses that should be excluded from the final accesslist
     excluded: HashSet<Address>,
     /// All addresses and touched slots
-    access_list: HashMap<Address, BTreeSet<B256>>,
+    touched_slots: HashMap<Address, BTreeSet<B256>>,
```
mattsse, line 29: this does make more sense

mattsse (approved):
makes sense

Issue #58, opened 2024-03-08
Add support for `JS` tracer to `muxTracer`


ArtificialPB:
Follow-up of https://github.com/paradigmxyz/evm-inspectors/pull/57, currently `muxTracer` supports only built-in tracers. It should also support `JS` tracers.

In the alloy-rs tracing crate, the `MuxConfig` and `MuxFrame` types will need to have map key changed to `GethDebugTracerType`. Changes in this project should be pretty much the same, with addition of supporting the `JSInspector` via `DelegatingInspector`. Changes in `reth` are probably not needed.

The `JSInspector` is under `js-tracer` feature, so a feature gate for it in `MuxInspector` will need to be added as well.

mattsse:
@ArtificialPB I'll need to bump evm-inspectors on reth shortly, so if you find the time I'd appreciate a pr that adds the missing `::Mux` handler for debug_trace* etc:

https://github.com/paradigmxyz/reth/blob/e21304947b6a46ba524d4b1e68351b05587d8892/crates/rpc/rpc/src/debug.rs#L279-L279

https://github.com/paradigmxyz/reth/blob/e21304947b6a46ba524d4b1e68351b05587d8892/crates/rpc/rpc/src/debug.rs#L512-L512

ArtificialPB:
@mattsse yeah, just opened it about an hour ago: https://github.com/paradigmxyz/reth/pull/7053

mattsse:
ah amazing!
ty

kustrun:
@mattsse @ArtificialPB Opened 2 PRs to support `JS tracer` in the `muxTracer`

Alloy: https://github.com/alloy-rs/alloy/pull/2442
Revm-inspectors (dependent on Alloy changes): https://github.com/paradigmxyz/revm-inspectors/pull/291

What’s the best way to test these changes in `reth`, given that using local path overrides for dependencies results in `error[E0308]: mismatched types` during compilation?

PR #291, opened 2025-05-13
feat(mux-tracer): Support JS tracer in MUX tracer.

kustrun:
Ref: https://github.com/paradigmxyz/revm-inspectors/issues/58

## Motivation

Currently `muxTracer` supports only built-in tracers. It should also support `JS tracers`.

Dependent on: https://github.com/alloy-rs/alloy/pull/2442

### DIFF
--- a/src/tracing/mux.rs
+++ b/src/tracing/mux.rs
@@ -7,6 +7,10 @@ use alloy_rpc_types_trace::geth::{
     CallConfig, FlatCallConfig, FourByteFrame, GethDebugBuiltInTracerType, NoopFrame,
     PreStateConfig,
 };
+#[cfg(feature = "js-tracer")]
+use alloy_rpc_types_trace::geth::{GethDebugTracerConfig, GethDebugTracerType, GethTrace};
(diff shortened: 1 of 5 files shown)


src/tracing/mux.rs
```diff
@@ -21,10 +25,13 @@ use revm::{
 use thiserror::Error;
 
 /// Mux tracing inspector that runs and collects results of multiple inspectors at once.
-#[derive(Clone, Debug)]
+#[derive(Debug)]
+#[cfg(feature = "js-tracer")]
```
mattsse, line 29: since the fields are private we can simply feature gate the addiiotnal jsInspector here

mattsse (changes requested):
cool, 

for testing you can patch the alloy deps to you branch

I'd still like to keep mux support without the json feature so we need sprinkle some features across this file

kustrun:
@mattsse As suggested, I have added the conditional support for JS tracer in MUX tracer.

DaniPopes:
stale

Issue #294, opened 2025-05-16
use correct from/to addresses for delegate calls in js tracer


mattsse:
correct geth behaviour
```json
            {
                "type": "CALL",
                "from": "0x0000000071727de22e5e9d8baf0edac6f37da032",
                "to": "0xabfa506cd083dc6e9bdd33b5cef847beaf940634",
                "method": "0x19822f7c",
                "gas": 49913,
                "value": "0"
            },
            {
                "type": "DELEGATECALL",
                "from": "0xabfa506cd083dc6e9bdd33b5cef847beaf940634",
                "to": "0x8e8e658e22b12ada97b402ff0b044d6a325013c7",
                "method": "0x19822f7c",
                "gas": 44301,
                "value": "0"
            },
            {
                "type": "STATICCALL",
                "from": "0xabfa506cd083dc6e9bdd33b5cef847beaf940634",
                "to": "0x0000000000000000000000000000000000000001",
                "method": "0xd4380c81",
                "gas": 41896
            },

```

this impl:

https://github.com/paradigmxyz/revm-inspectors/blob/edf4eb534aa8deade9e72a0652aae135b0aba717/src/tracing/js/mod.rs#L473-L477

```json
            {
                "from": "0x0000000071727de22e5e9d8baf0edac6f37da032",
                "gas": 49913,
                "method": "0x19822f7c",
                "to": "0xabfa506cd083dc6e9bdd33b5cef847beaf940634",
                "type": "CALL",
                "value": "0"
            },
            {
                "from": "0x0000000071727de22e5e9d8baf0edac6f37da032",
                "gas": 44301,
                "method": "0x19822f7c",
                "to": "0xabfa506cd083dc6e9bdd33b5cef847beaf940634",
                "type": "DELEGATECALL",
                "value": "0"
            },
```

ref https://github.com/paradigmxyz/reth/issues/16289

we have this for the regular tracer:

https://github.com/paradigmxyz/revm-inspectors/blob/edf4eb534aa8deade9e72a0652aae135b0aba717/src/tracing/mod.rs#L568-L574

geth ref:

https://github.com/ethereum/go-ethereum/blob/1f175347c47810564c7d6fab3bbd855eb449468d/eth/tracers/js/goja.go#L386-L386

PR #295, opened 2025-05-16
fix: js tracer behavior

klkvr:
Closes https://github.com/paradigmxyz/revm-inspectors/issues/294 while retaining fix for https://github.com/paradigmxyz/reth/issues/13089

The idea is that in steps (aka logs) we want `contract.address` to always be the address of contract which storage is accessed (`target_address`). However, when dealing with delegatecall frames, `contract.address` should be set to the `bytecode_address` as it is marking the destination of delegatecall.

This means we can't just forward `active_call().contract` to steps anymore and need to re-construct the `Contract` value manually

### DIFF
--- a/src/tracing/js/mod.rs
+++ b/src/tracing/js/mod.rs
@@ -418,6 +418,7 @@ where
         let (stack, _stack_guard) = StackRef::new(&interp.stack);
         let evm_memory = interp.memory.borrow();
         let (memory, _memory_guard) = MemoryRef::new(evm_memory);
+        let active_call = self.active_call();
         let step = StepLog {
             stack,
             op: interp.bytecode.opcode().into(),
@@ -431,8 +432,8 @@ where
             contract: Contract {
                 caller: interp.input.caller_address,
                 contract: interp.input.target_address,
-                value: self.active_call().contract.value,
-                input: self.active_call().contract.input.clone(),
+                value: active_call.contract.value,
+                input: active_call.contract.input.clone(),
             },
         };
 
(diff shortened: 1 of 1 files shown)


src/tracing/js/mod.rs
```diff
@@ -428,7 +428,12 @@ where
             depth: context.journal_ref().depth() as u64,
             refund: interp.control.gas().refunded() as u64,
             error: None,
-            contract: self.active_call().contract.clone(),
+            contract: Contract {
+                caller: interp.input.caller_address,
+                contract: interp.input.target_address,
+                value: self.active_call().contract.value,
+                input: self.active_call().contract.input.clone(),
```
mattsse, line 435: we can get rid of one  active_call here, because all this does is .last() but still

src/tracing/js/mod.rs
```diff
@@ -458,7 +463,12 @@ where
                 depth: context.journal_ref().depth() as u64,
                 refund: interp.control.gas().refunded() as u64,
                 error: Some(format!("{:?}", interp.control.instruction_result())),
-                contract: self.active_call().contract.clone(),
+                contract: Contract {
+                    caller: interp.input.caller_address,
+                    contract: interp.input.target_address,
+                    value: self.active_call().contract.value,
+                    input: self.active_call().contract.input.clone(),
```
mattsse, line 470: same here

mattsse (changes requested):
smol pedantic nit

mattsse (approved):

PR #299, opened 2025-05-22
chore: Remove eof trace handlers

### LINKED ISSUE #298: Remove eof trace handlers
https://github.com/paradigmxyz/revm-inspectors/blob/0814a4d44525f63e4c111826af11955b075eb851/src/tracing/mod.rs#L638-L638

we can remove all of the eof handler overrides here

## TODO
* remove all eof handler implementations

stevencartavia:
closes #298

### DIFF
--- a/src/access_list.rs
+++ b/src/access_list.rs
@@ -172,16 +172,4 @@ where
         }
         None
     }
-
-    fn eofcreate(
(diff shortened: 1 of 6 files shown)


mattsse (approved):
ty champ

PR #309, opened 2025-07-03
fix: use revert directly

mattsse:


rakita (approved):
REVERT is the only available option here. lgtm

PR #311, opened 2025-07-03
fix: always record revert

mattsse:


src/tracing/js/mod.rs
```diff
@@ -413,6 +413,12 @@ where
     CTX: ContextTr<Journal: JournalExt, Db: DatabaseRef>,
 {
     fn step(&mut self, interp: &mut Interpreter, context: &mut CTX) {
+        if interp.bytecode.opcode() == OpCode::REVERT.get() {
+            // if this is a revert we need to manually record this so that we can use it in the
+            // step_end fn
+            self.revert_pc = Some(interp.bytecode.pc());
+        }
```
rakita, line 420: In general, don't need the check for revert, you can assign the pc every time

mattsse: makes sense, let me simplify

rakita (approved):

PR #316, opened 2025-07-11
feat: geth_erc7562_tracers addition

### LINKED ISSUE #315: Add support for erc-7562 frame
ref https://github.com/alloy-rs/alloy/pull/2682

once released we need another gethbuilder function that populates this from the recorded nodes, similar to other frames, like

https://github.com/paradigmxyz/revm-inspectors/blob/eb6ed5965050e8723ccf1ec02b0b53f7f0005af5/src/tracing/builder/geth.rs#L220

or 

https://github.com/paradigmxyz/revm-inspectors/blob/eb6ed5965050e8723ccf1ec02b0b53f7f0005af5/src/tracing/builder/geth.rs#L135

Rimeeeeee:
fixes #315

### DIFF
--- a/src/tracing/builder/geth.rs
+++ b/src/tracing/builder/geth.rs
@@ -10,6 +10,7 @@ use alloc::{
 };
 use alloy_primitives::{map::HashMap, Address, Bytes, B256, U256};
 use alloy_rpc_types_trace::geth::{
+    erc7562::{AccessedSlots, ContractSize, Erc7562Config, Erc7562Frame},
     AccountChangeKind, AccountState, CallConfig, CallFrame, DefaultFrame, DiffMode,
     GethDefaultTracingOptions, PreStateConfig, PreStateFrame, PreStateMode, StructLog,
 };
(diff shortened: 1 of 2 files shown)


src/tracing/builder/geth.rs
```diff
@@ -347,4 +353,195 @@ impl<'a> GethTraceBuilder<'a> {
...
+            // EXT opcode handling and contract size capture
+            if matches!(op, opcode::EXTCODESIZE | opcode::EXTCODECOPY | opcode::EXTCODEHASH) {
+                if let Some(stack) = &step.stack {
+                    if let Some(item) = stack.get(stack.len().saturating_sub(1)) {
+                        let address = Address::from(item.to_be_bytes());
+                        ext_code_access_info.push(format!("{address:?}"));
+                        if let Entry::Vacant(e) = contract_size.entry(address) {
+                            if let Ok(Some(account)) = db.basic_ref(address) {
+                                if let Ok(bytecode) = db.code_by_hash_ref(account.code_hash) {
+                                    let size = bytecode.len();
```
mattsse, line 469: I think we should check if the hash is keccak_empty just to be sure

src/tracing/builder/geth.rs
```diff
@@ -347,4 +353,195 @@ impl<'a> GethTraceBuilder<'a> {
...
+                    }
+                }
+            }
+        }
+
+        let mut calls = Vec::new();
+        for (idx, n) in self.nodes.iter().enumerate() {
+            if n.parent == Some(node_idx) {
+                let child_trace =
+                    self.geth_erc7562_traces_internal(idx, opts.clone(), gas_used, &db);
```
mattsse, line 505: this is a recursive call, right?

this tree can be quite deep and can result in stack overflows, so we should solve this with a stack based approach instead, similar to callframe impl

src/tracing/builder/geth.rs
```diff
@@ -347,4 +353,195 @@ impl<'a> GethTraceBuilder<'a> {
...
+    /// Converts a CallKind to a CallFrameType.
+    pub fn convert_call_kind(kind: CallKind) -> CallFrameType {
+        match kind {
+            CallKind::Call => CallFrameType::Call,
+            CallKind::CallCode => CallFrameType::CallCode,
+            CallKind::DelegateCall => CallFrameType::DelegateCall,
+            CallKind::StaticCall => CallFrameType::StaticCall,
+            CallKind::Create => CallFrameType::Create,
+            CallKind::Create2 => CallFrameType::Create2,
+            CallKind::AuthCall => todo!(),
```
mattsse, line 544: I think this we can just map to `Call`

mattsse (changes requested):
very nice, only concern is the recursive nature of the _inner call but we should be able to convert this to an iterative stack based variant instead

src/tracing/builder/geth.rs
```diff
@@ -347,4 +354,199 @@ impl<'a> GethTraceBuilder<'a> {
...
+                return frame;
+            }
+        }
+
+        // Fallback in case no root found, to satisfy the compiler
+        frames
+            .into_iter()
+            .find(|(idx, _)| self.nodes[*idx].parent.is_none())
+            .map(|(_, frame)| frame)
+            .unwrap_or_default()
```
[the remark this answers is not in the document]
Rimeeeeee: clean :)

mattsse (approved):
great!

PR #323, opened 2025-07-19
perf: allocate some more initial capacity for CallTraceArena

DaniPopes:
The `vec![]` macro only allocates for the number of elements. In `CallTraceArena::default()`, it only allocates 1, so when growing it the capacity doubles starting at 1, which is pretty inefficient.


mattsse (approved):
makes sense

PR #325, opened 2025-07-20
feat: reused call trace stack

### LINKED ISSUE #321: Reuse call trace stack
when we fuse the inspector we clear all contents of

https://github.com/paradigmxyz/revm-inspectors/blob/0f02549ecaa2e0dfdf13e328d2089276179d9457/src/tracing/mod.rs#L114-L114

this includes the rather large steps stack:

https://github.com/paradigmxyz/revm-inspectors/blob/0f02549ecaa2e0dfdf13e328d2089276179d9457/src/tracing/mod.rs#L449-L449

ideally we want only want to clear those but reuse them across multiple runs. for this we can park the cleared step stacks them in another vec when we fuse the inspector, and when we initialize a new trace we pop from this or create a new vec

Rimeeeeee:
fixes #321

### DIFF
--- a/src/tracing/mod.rs
+++ b/src/tracing/mod.rs
@@ -85,6 +85,9 @@ pub struct TracingInspector {
     ///
     /// This is filled during execution.
     spec_id: Option<SpecId>,
+    /// Pool of reusable step vectors to reduce allocations
+    reusable_step_vecs: Vec<Vec<CallTraceStep>>,
+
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
cool, needs smol rebase and some suggestions

mattsse (approved):

PR #326, opened 2025-07-20
feat: boxed the decoded field

### LINKED ISSUE #324: Box the `decoded` fields
The `decoded` fields are used only for printing the trace which is not the common path; we can use `Option<Box` to reduce memory usage by reducing the size of the trace structs

Rimeeeeee:
fixes #324


src/tracing/types.rs
```diff
@@ -96,7 +97,7 @@ pub struct CallTrace {
     /// Opcode-level execution steps.
     pub steps: Vec<CallTraceStep>,
     /// Optional complementary decoded call data.
-    pub decoded: DecodedCallTrace,
+    pub decoded: Option<Box<DecodedCallTrace>>,
```
mattsse, line 100: should we also make this an option @DaniPopes ?

I guess so because then we dont need to box two options

src/tracing/writer.rs
```diff
@@ -265,32 +265,38 @@ impl<W: Write> TraceWriter<W> {
...
-                    (name.to_string(), args.join(", "))
-                }
-                None => {
-                    if trace.data.len() < 4 {
-                        ("fallback".to_string(), hex::encode(&trace.data))
-                    } else {
-                        let (selector, data) = trace.data.split_at(4);
-                        (hex::encode(selector), hex::encode(data))
+            let (func_name, inputs) =
+                match &trace.decoded.as_ref().and_then(|d| d.call_data.as_ref()) {
```
mattsse, line 276: I think we can introduce some helper functions to access these fields than this becomes less verbose and can get rid of a bunch of nested and_then asref

src/tracing/writer.rs
```diff
@@ -319,9 +325,9 @@ impl<W: Write> TraceWriter<W> {
         let log_style = self.log_style();
         self.write_branch()?;
 
-        if let Some(name) = &log.decoded.name {
+        if let Some(name) = &log.decoded.as_ref().unwrap().name {
             write!(self.writer, "emit {name}({log_style}")?;
-            if let Some(params) = &log.decoded.params {
+            if let Some(params) = &log.decoded.as_ref().unwrap().params {
```
mattsse, line 330: here we can also use a new helper fn that returns an option, then we can avoid the unwrap

tests/it/writer/deploy_fail/decoded.storage.txt
```diff
@@ -1,2 +1,2 @@
-  [18] → new RevertingConstructor@0xBd770416a3345F91E4B34576cb804a576fa48EB1
-    └─ ← [Revert] 42
+  [18] → new <unknown>@0xBd770416a3345F91E4B34576cb804a576fa48EB1
+    └─ ← [Revert] 0x0000000000000000000000000000000000000000000000000000000000000042
```
mattsse, line 2: hmm, should this change?

this likely indicates a current regression where we fail to record the decoded value perhaps?

mattsse (changes requested):
great start

src/tracing/writer.rs
```diff
@@ -555,3 +557,19 @@ fn num_or_hex(x: U256) -> String {
...
+    trace.decoded.as_ref()?.call_data.as_ref()
+}
+
+fn decoded_return_data(trace: &CallTrace) -> Option<&str> {
+    trace.decoded.as_ref()?.return_data.as_deref()
+}
+
+fn decoded_log(log: &CallLog) -> &DecodedCallLog {
+    log.decoded.as_ref().unwrap()
+}
```
mattsse, line 575: I think we can make these as pub(crate) functions of struct CallTrace directly

mattsse (changes requested):

src/tracing/types.rs
```diff
@@ -166,7 +167,7 @@ impl From<Log> for CallLog {
         Self {
             position: Default::default(),
             raw_log: log.data,
-            decoded: DecodedCallLog { name: None, params: None },
+            decoded: Some(Box::new(DecodedCallLog { name: None, params: None })),
```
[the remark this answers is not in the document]
Rimeeeeee: if this is made directly none, it cant unwrap for test `test_tracing_printing()`

DaniPopes (approved):
Thanks!

PR #345, opened 2025-09-02
chore: remove debug

DaniPopes:
Default hash builder implements Debug conditionally


mattsse:
CI unrelated, had to yank alloy with secp 31 -.-
will do first thing

PR #351, opened 2025-09-19
perf: re-use returndata buffer

DaniPopes:
`return_data.buffer()` is `Bytes`. Cheaply clone it instead of creating a new copy.


mattsse (approved):
hell yeah

PR #353, opened 2025-09-21
fix(tracing): convert empty revertReason to none

jsvisa:
Geth will convert the empty bytes to null, and omit it in the response, I think we should align with it, the empty nulls seems meanless.

Below is the differ of the tx:  https://etherscan.io/tx/0x337749b7c19859b2d773868b60e82a06ae008db9a80d41db76d7020d976a72e7

> left is geth, right is reth

image

### DIFF
--- a/src/tracing/utils.rs
+++ b/src/tracing/utils.rs
@@ -123,6 +123,8 @@ pub(crate) fn maybe_revert_reason(output: &[u8]) -> Option<String> {
             let reason = match err {
                 // return the raw revert reason and don't use the revert's display message
                 ContractError::Revert(revert) => revert.reason,
+                // return as geth panic error string
+                ContractError::Panic(panic) => panic.as_geth_str().to_string(),
                 err => err.to_string(),
             };
 


src/tracing/utils.rs
```diff
@@ -126,7 +126,7 @@ pub(crate) fn maybe_revert_reason(output: &[u8]) -> Option<String> {
         }
         GenericRevertReason::RawString(err) => err,
     };
-    if reason.is_empty() {
+    if reason.is_empty() || reason.trim_matches('\0').is_empty() {
```
mattsse, line 129: can we document this here

trying to find the geth equivalent for this, and unsure where the equivalent check for this is

maybe this here?

https://github.com/ethereum/go-ethereum/blob/85077be58edea572f29c3b1a6a055077f1a56a8b/accounts/abi/abi.go#L278-L281

mattsse (changes requested):
ty,

smol doc nit and q

jsvisa:
Copied the revert/error handling from geth's, PTAL.

BTW, also changed the panic message in https://github.com/alloy-rs/core/pull/1015

src/tracing/utils.rs
```diff
@@ -113,23 +113,44 @@ pub(crate) fn load_account_code<DB: DatabaseRef>(
...
+            _ => None,
+        })
+    } else if output[0..4] == PANIC_SELECTOR {
+        // Try to decode Panic(uint256)
+        GenericRevertReason::decode(output).and_then(|reason| match reason {
+            GenericRevertReason::ContractError(ContractError::Panic(panic)) => {
+                Some(panic.to_string())
+            }
+            _ => None,
+        })
```
mattsse, line 150: @DaniPopes isnt this already part of the GenericRevertReason decode impl?

src/tracing/utils.rs
```diff
@@ -113,23 +113,44 @@ pub(crate) fn load_account_code<DB: DatabaseRef>(
...
+            GenericRevertReason::ContractError(ContractError::Revert(revert)) => {
+                let reason = revert.reason;
+                if reason.is_empty() || reason.trim_matches('\0').is_empty() {
+                    None
+                } else {
+                    Some(reason)
+                }
             }
-        }
-        GenericRevertReason::RawString(err) => err,
```
mattsse, line 127: i think we only need to add the empty check here?

mattsse (changes requested):

src/tracing/utils.rs
jsvisa: thanks for the reminder, seems we already have the decode logic https://github.com/alloy-rs/core/blob/9bdf3c764793c68acbee88a6f7c40574ae86a63b/crates/sol-types/src/types/interface/mod.rs#L286-L293

mattsse:
hmm, @alpergundogdu 

I wasnt able to reproduce this with this test https://github.com/paradigmxyz/revm-inspectors/pull/386

PR #357, opened 2025-09-25
perf: remove step_stack

DaniPopes:
This stack should always be exactly 1 element, and the context should be the same in both `step` and `step_end`.

This avoids unnecessary memory bandwidth and `arena` access in case of `!record` in the very hot `step` functions.

### DIFF
--- a/src/tracing/mod.rs
+++ b/src/tracing/mod.rs
@@ -75,8 +75,8 @@ pub struct TracingInspector {
     traces: CallTraceArena,
     /// Tracks active calls
     trace_stack: Vec<usize>,
-    /// Tracks active steps
-    step_stack: Vec<StackStep>,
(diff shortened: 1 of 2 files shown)


src/tracing/mod.rs
```diff
@@ -473,26 +468,31 @@ impl TracingInspector {
 
         self.last_journal_len = context.journal_ref().journal().len();
 
-        trace.trace.steps.push(CallTraceStep {
+        let step_idx = node.trace.steps.len();
+        node.trace.steps.push(CallTraceStep {
+            depth: context.journal().depth() as u64,
             pc: interp.bytecode.pc(),
             op,
+            contract: interp.input.target_address(),
```
mattsse, line 476: are we adding those back now?

mattsse (commented):
I believe this is indeed only one element always so we can get rid of this

PR #358, opened 2025-09-25
chore: fix ensure `$root` is always an absolute path

mdqst:
added `realpath` when calculating the root, so `$root` will always be an absolute path, no matter where or how the script is run.

p.s. this prevents errors with `git cliff --workdir` when the script is executed from a different directory.


mattsse (commented):
this has been working fine

PR #359, opened 2025-09-29
Fix return data in geth traces

pgherveou:
The Geth trace builder incorrectly sets the same `returnData` value for all opcode steps in a trace, 
using the final call output instead of the step-specific return data buffer state.

## Expected Behavior

Each opcode step should display the return data buffer state
at that specific point in execution.

## Actual Behavior

All opcode steps show the same returnData value - the final output of the entire
call. 

For example, in a Fibonacci contract call that returns 1, every single opcode
step shows:
"returnData": "0x0000000000000000000000000000000000000000000000000000000000000001"


pgherveou:
🙏  you think you could push a patch release on crates.io that includes this?

PR #382, opened 2025-11-13
fix: dont filter out created+selfdestruct in pre

mattsse:
should close this one https://github.com/paradigmxyz/reth/issues/19703

previously we were yeeting them from pre here:

https://github.com/paradigmxyz/revm-inspectors/blob/cbf97c9947f02e07dbe0c5e27e32fbb3d205114d/src/tracing/builder/geth.rs#L378-L379

### DIFF
--- a/src/tracing/builder/geth.rs
+++ b/src/tracing/builder/geth.rs
@@ -319,7 +319,15 @@ impl<'a> GethTraceBuilder<'a> {
 
             // determine the change type
             let pre_change = if changed_acc.is_created() {
-                AccountChangeKind::Create
+                // if the account was created _and_ selfdestructed we need to treat it as modified,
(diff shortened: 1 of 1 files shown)


antoniolocascio:
I still have a question about the line:
https://github.com/paradigmxyz/revm-inspectors/blob/cbf97c9947f02e07dbe0c5e27e32fbb3d205114d/src/tracing/builder/geth.rs#L378-L379

A contract can be deployed at an address with non-zero balance. If that happens (and it doesn't selfdestruct), the pre section should contain the address with only balance. IIUC, this line would still filter it out. Basically, the rule should be to filter out from pre contracts that have been deployed on empty addresses. I don't have an example to test on geth at hand, but should be easy enough to find one.

PR #385, opened 2025-12-10
feat: stage revm to 34.0.0

rakita:
stage latest main commit


gakonst (commented):
also enable gmp feature

onbjerg (commented):
pls remove the `mod.rs:<line>:<col>` files

Cargo.toml
```diff
@@ -65,6 +65,3 @@ std = [
 ]
 serde = ["dep:serde", "revm/serde"]
 js-tracer = ["dep:boa_engine", "dep:boa_gc"]
-
-[patch.crates-io]
-#revm = { git = "https://github.com/bluealloy/revm", rev = "e13b609fbbc80f9e5858255e4412db95c0cc6d12" }
```
mattsse: we can keep these in the future because this is convenient

PR #388, opened 2025-12-16
feat(TraceWriter): order the storage writes in last-write order

0xferrous:
The storage writes were written in non determinstic order because of HashMap usage. Instead, to make it more intuitive, the storage slot changes are shown in the order of their last writes.


0xferrous:
I personally think intermediate storage writes can be useful too when debugging. Perhaps in another level deep of verbosity.

mattsse (commented):
dont really have an opinion here

defer to @DaniPopes

DaniPopes (approved):
lgtm

PR #390, opened 2025-12-29
feat: add features for skipping ser/de of call trace steps

0xferrous:
Serializing `CallTraceStep`s is too verbose, one `CallTraceArean` for a single trace of a heavy transaction can lead to huge data.

E.g.: A GMX order execution transaction's `CallTraceArena` with steps exceeds 2GB(without steps it is 0.6mb) when serialized to JSON. And most of that is because of the steps getting serialized too.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,12 @@ std = [
     "thiserror/std",
 ]
 serde = ["dep:serde", "revm/serde"]
+serde-skip-trace-steps = [
+    "skip-trace-steps-ser",
+    "skip-trace-steps-deser",
+]
+skip-trace-steps-ser = []
+skip-trace-steps-deser = []
 js-tracer = ["dep:boa_engine", "dep:boa_gc"]
 
 [patch.crates-io]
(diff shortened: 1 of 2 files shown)


0xferrous:
I am writing a TUI for interactive foundry traces(folding, showing current call's args, return data in a separate column, copying current calldata/returndata, etc.). Writing the trace will require me to have all the trace items in memory since that's how the graph is laid out, so I can't iterate through the `CallTraceNode`s in a consume and forget kind of manner.

Even if I use `serde_json::StreamDeserializer` it'll still continue to spend time on deserializing all of the `CallTraceStep`s instead of skipping over them.

FWIW I tried to decode the items with steps skipped, it cut the time to deserialize from 2 minutes to 1 minute, but even that is a lot. So, for me the solution is to not have it serialized into the json at all. I am still trying to figure out how to ignore steps if it already exists in the json in a manner where it won't take so long to decode it.

But generally it seems logical to allow this option, no? Since it encodes every single opcode step data and it's only useful for a debugger. I think best solution is to avoid json completely, but right now that is the only way for me to do it outside foundry as `cast run <tx> --json | tui`.

0xferrous:
I think I am gonna ignore this problem for now and see if I can think of the solution to this passively.

PR #396, opened 2026-01-20
fix(js): correct gas cost reporting by invoking step callback in step_end

### LINKED ISSUE #200: Fix gas cost for js tracer

### LINKED ISSUE #10959: (Custom EVM tracer): Reported opcode gas cost is incorrect
### Describe the bug

When using a custom evm tracer, the gas cost for opcoded reported by the tracer is incorrect.

### Steps to reproduce

Consider the following `debug_traceTransaction` RPC reqeust:

```python
import asyncio
import json
import os

from aiohttp import ClientSession, TCPConnector


async def custom_tracer():
    async with ClientSession(connector=TCPConnector(limit=0)) as session:
        method = "debug_traceTransaction"
        tracer = """
        {
            data: [],
            memoryInstructions: {"MSTORE":"W", "MSTORE8":"B", "MLOAD":"R"},
            fault: function (log) {},
            step: function (log) {
                let op = log.op.toString();
                let instructions = this.memoryInstructions
                if (Object.keys(instructions).includes(op)) {
                this.data.push({
                    op: instructions[op] ,
                    depth: log.getDepth(),
                    offset: log.stack.peek(0),
                    gasCost: log.getCost(),
                    memorySize: log.memory.length(),
                });
                }
            },
            result: function (ctx, _) {
                return {
                    error: !!ctx.error,
                    data: this.data
                };
            }
        }
        """
        params = [
            "0x82f6413a2658ebb83f27e44fe1e815ec0979a7dcc0bc9dbdbbbe058d547195a6",
            {"tracer": tracer},
        ]

        payload = {
            "jsonrpc": "2.0",
            "method": method,
            "params": params,
            "id": 1,
        }

        async with session.post(
            "https://lb.nodies.app/v2/8bfbba93396a46b1a3d99311daf0dae6",
            json=payload,
            headers={"x-api-key": "b27881a7-4ea2-439f-a1c6-77ad0c875c16"},
        ) as response:
            result = await response.json()
            if "error" in result:
                raise Exception(f"RPC Error ({method}>req:{params}): {result['error']}")

            with open(os.path.join("trace.json"), "w") as output_file:
                json.dump(result, output_file)
asyncio.run(custom_tracer())
```
Expected gas costs:

```json
[
  6,  6, 3, 3, 12, 3,  3,  9, 6, 3, 3, 3,
  3,  3, 3, 3,  3, 6,  6,  3, 3, 3, 3, 3,
  3,  3, 3, 3,  3, 3,  3,  3, 3, 6, 6, 3,
  3,  3, 3, 6,  6, 3,  3,  3, 3, 3, 3, 6,
  6,  3, 6, 3,  3, 3,  3, 12, 3, 3, 9, 6,
  3, 12, 6, 3,  3, 3,  3,  3, 3, 3, 3, 6,
  3,  6, 3, 6,  3, 3, 12,  3, 3, 3, 3, 3,
  3,  3, 6, 3,  3, 3,  3,  3, 3, 3, 3, 3,
  3,  3, 3, 3,
  ... 5998 more items
]
```

Reported gas cost:
```json
[
    138,   150,   229,   317,     6,  2148,  2163,  2175,  2196,
   2642,  2657,  2925,  2940,  3191,  3206,  4376,  4391,  4403,
   4421,  4746,  4761,  4996,  5011,  5284,  5299,  5875,  5890,
   6643,  6658,  7284,  7299,  7493,  7508,  7520,  7538,  7867,
   7882, 10896, 10911, 10923, 10941, 11270, 11285, 11741, 11756,
  14788, 14803, 14815, 14833, 15142, 15256, 15297, 17115, 17235,
  17273,     6,  4184,  4199,  4208,  4232,  4241,  4265,  4295,
   4304,  4328,  4346,  4361,  4379,  4385,  4391,  4465,  4471,
   4547,  4553,  4714,  4726,  4747,  4783,     6,   415,   430,
    697,   712,  3820,  3835,  6943,  6955,  6977,  8804,  8828,
   8847, 13805, 13839, 13847, 13921, 13960, 13991, 14054, 14063,
  14077,
  ... 5998 more items
]
```

### Node logs

_No response_

### Platform(s)

Linux (x86)

### What version/commit are you on?

reth Version: 1.0.0-dev
Commit SHA: 4fd06bd3b
Build Timestamp: 2024-07-06T20:32:43.565379262Z
Build Features: jemalloc
Build Profile: maxperf

### What database version are you on?

Current database version: 2
Local database is uninitialized

### Which chain / network are you on?

mainnet

### What type of node are you running?

Archive (default)

### What prune config do you use, if any?

_No response_

### If you've built Reth from source, provide the full command you used

_No response_

### Code of Conduct

- [X] I agree to follow the Code of Conduct

mattsse:
## Summary

This fixes the long-standing issue where `log.getCost()` in JS tracers returned incorrect gas costs (paradigmxyz/reth#10959, #200).

## Problem

The JS step callback was invoked in `step()` (before opcode execution), but the gas cost is only known after the opcode executes. This caused the first opcode to report cost=0 and each subsequent opcode to report the *previous* opcode's cost.

**Before (incorrect)**: `[0, 3, 3]` for PUSH1, PUSH1, STOP
**After (correct)**: `[3, 3, 0]` for PUSH1, PUSH1, STOP

## Solution

The key insight is that we need the **pre-execution stack/memory state** but the **post-execution gas cost**. This requires a two-phase approach:

1. `step()`: Capture pre-execution snapshot (stack clone, memory snapshot, pc, op, etc.)
2. `step_end()`: Compute gas delta and invoke JS callback with the snapshot

### Changes

- Add `PendingStep` struct to capture pre-execution state in `step()`
- Add `MemorySnapshot` to properly snapshot memory contents (SharedMemory clone is shallow due to Rc)
- Invoke the JS step callback in `step_end()` with captured state and correct gas cost
- Update tests to expect correct gas costs

## Testing

Verified against live node that the fix produces correct gas costs:

```
curl -X POST http://node:8545 -d '{
  "method": "debug_traceTransaction",
  "params": ["0x...", {"tracer": "{ ... log.getCost() ... }"}]
}'
```

All existing tests pass.

## Breaking Change Note

JS tracer errors that occur on the final opcode (like STOP) can no longer revert the transaction since the opcode has already completed. The transaction succeeds but the trace may be incomplete. This matches the expected behavior since the tracer is observational.

Fixes: paradigmxyz/reth#10959
Closes: #200

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,8 +30,8 @@ clippy.lint_groups_priority = "allow"
 
 [dependencies]
 # eth
-alloy-rpc-types-eth = { version = "1.0.38", default-features = false }
-alloy-rpc-types-trace = { version = "1.0.38", default-features = false }
+alloy-rpc-types-eth = { version = "1.4.3", default-features = false }
+alloy-rpc-types-trace = { version = "1.4.3", default-features = false }
 alloy-sol-types = { version = "1.4", default-features = false }
 alloy-primitives = { version = "1.4", default-features = false, features = [
     "map",
--- a/src/tracing/js/bindings.rs
+++ b/src/tracing/js/bindings.rs
@@ -13,6 +13,7 @@ use alloc::{
     format,
     rc::Rc,
     string::{String, ToString},
+    vec::Vec,
 };
 use alloy_primitives::{Address, Bytes, B256, U256};
 use boa_engine::{
@@ -236,14 +237,33 @@ impl StepLog {
     }
 }
 
+/// A snapshot of memory contents as a `Vec<u8>`
+#[derive(Clone, Debug, Default)]
+pub(crate) struct MemorySnapshot(pub(crate) Vec<u8>);
+
+impl MemorySnapshot {
+    /// Create a snapshot from SharedMemory
+    pub(crate) fn from_shared_memory(mem: &SharedMemory) -> Self {
+        Self(mem.context_memory().to_vec())
+    }
+
+    fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    fn slice_len(&self, offset: usize, size: usize) -> &[u8] {
+        &self.0[offset..offset + size]
+    }
+}
+
 /// Represents the memory object
 #[derive(Clone, Debug)]
-pub(crate) struct MemoryRef(GuardedNullableGc<SharedMemory>);
+pub(crate) struct MemoryRef(GuardedNullableGc<MemorySnapshot>);
 
 impl MemoryRef {
-    /// Creates a new stack reference
-    pub(crate) fn new(mem: &SharedMemory) -> (Self, GcGuard<'_, SharedMemory>) {
-        let (inner, guard) = GuardedNullableGc::new_ref(mem);
+    /// Creates a new memory reference from an owned snapshot (for deferred step invocation)
+    pub(crate) fn new_owned(mem: MemorySnapshot) -> (Self, GcGuard<'static, MemorySnapshot>) {
+        let (inner, guard) = GuardedNullableGc::new_owned(mem);
         (Self(inner), guard)
     }
 
@@ -430,9 +450,9 @@ impl From<u8> for OpObj {
 pub(crate) struct StackRef(GuardedNullableGc<Stack>);
 
 impl StackRef {
-    /// Creates a new stack reference
-    pub(crate) fn new(stack: &Stack) -> (Self, GcGuard<'_, Stack>) {
-        let (inner, guard) = GuardedNullableGc::new_ref(stack);
+    /// Creates a new stack reference from an owned stack (for deferred step invocation)
+    pub(crate) fn new_owned(stack: Stack) -> (Self, GcGuard<'static, Stack>) {
+        let (inner, guard) = GuardedNullableGc::new_owned(stack);
         (Self(inner), guard)
     }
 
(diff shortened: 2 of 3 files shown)


yongkangc (approved):
This is an interesting approach, makes sense

DaniPopes:
superseded by https://github.com/paradigmxyz/revm-inspectors/pull/423

PR #398, opened 2026-01-22
feat: add TimeoutInspector for execution time limits and cancellation

gakonst:
## Summary

Adds timeout and cancellation support directly to `TracingInspector` and `JsInspector` via an optional `TimeoutConfig`:

- **Duration-based timeouts** (requires `std` feature)
- **External cancellation** via `AtomicBool` signal (useful when requests are dropped/cancelled)
- **Configurable opcode check interval** - check every N opcodes instead of only at call boundaries
- **no_std support** for cancellation-only mode

## Motivation

The evm-inspectors implementation doesn't come with timeout support, so execution can effectively run forever if certain overrides are provided (such as max gaslimit and price 0).

This is inspired by the [`TimeLimit` inspector in trevm](https://github.com/init4tech/trevm/blob/main/src/inspectors/timeout.rs) but integrated directly into the existing inspectors for a cleaner API.

## Usage

```rust
use revm_inspectors::timeout::TimeoutConfig;
use revm_inspectors::tracing::{TracingInspector, TracingInspectorConfig};
use std::{time::Duration, sync::{Arc, atomic::AtomicBool}};

// Duration timeout with interval checking every 1000 opcodes
let timeout = TimeoutConfig::new(Duration::from_secs(5))
    .with_check_interval(1000);

let inspector = TracingInspector::new(TracingInspectorConfig::default_geth())
    .with_timeout(timeout);

// With external cancellation signal (for request dropping)
let cancel = Arc::new(AtomicBool::new(false));
let timeout = TimeoutConfig::new(Duration::from_secs(5))
    .with_signal(cancel.clone());

// From another task: cancel execution
cancel.store(true, Ordering::Relaxed);

// Cancellation-only mode (no_std compatible)
let timeout = TimeoutConfig::cancellation_only(cancel);
let inspector = TracingInspector::new(config).with_timeout(timeout);

// Also works with JsInspector
let js_inspector = JsInspector::new(code, config)?
    .with_timeout(timeout);
```

## Check Points

Timeout/cancellation is checked at:
- Call boundaries (`call`, `call_end`)
- Create boundaries (`create`, `create_end`)
- During step execution if `check_interval` is configured

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,4 +36,7 @@ pub mod edge_cov;
 /// An inspector for tracking storage access.
 pub mod storage;
 
+/// Timeout and cancellation inspectors for limiting EVM execution.
+pub mod timeout;
+
 pub use colorchoice::ColorChoice;
--- a/src/timeout.rs
+++ b/src/timeout.rs
@@ -1,112 +1,68 @@
-//! Configurable timeout inspector for limiting EVM execution time.
+//! Timeout configuration for limiting EVM execution time.
 //!
-//! This module provides a [`TimeoutInspector`] that can be used to limit the wall-clock time
-//! spent on EVM execution. It supports:
+//! This module provides [`TimeoutConfig`] for configuring execution time limits
+//! and external cancellation signals. It can be used with [`TracingInspector`] and
+//! [`JsInspector`] via their timeout configuration methods.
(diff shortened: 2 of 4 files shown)


DaniPopes:
stale

PR #404, opened 2026-02-12
feat(edge_cov): collision-free dense edge IDs

gakonst:
## Summary

Replace the hash-modulo scheme (`hash(addr,pc,dest) % 65536`) with a `HashMap`-based dense ID assignment that eliminates edge collisions in coverage-guided fuzzing.

## Motivation

The current `EdgeCovInspector` hashes each `(address, pc, jump_dest)` tuple and truncates to a 65536-entry buffer. With large contracts or instrumented native code (e.g. sancov-instrumented precompiles), distinct edges frequently alias to the same hitcount slot, corrupting the coverage signal. The fuzzer can't distinguish "new edge A" from "more hits on existing edge B", degrading guidance quality.

## Changes

- `EdgeCovInspector` now holds a `HashMap<(Address, usize, U256), usize>` mapping each unique edge to a dense monotonic ID
- New edges are assigned IDs on the cold path (first encounter); known edges hit the `Occupied` fast path — effectively the same cost as the previous hash, since both require hashing the same key
- Buffer pre-allocated to 65536 entries (configurable via `with_capacity()`); edges beyond capacity are silently dropped
- `get_hitcount()` returns only the used portion `[0..edge_count()]` instead of the full buffer
- `reset()` clears counters but preserves ID assignments for reuse across iterations
- Hitcount uses `saturating_add` instead of `checked_add().unwrap_or()`

New public API (backward-compatible additions):
- `with_capacity(n)` — size the buffer for workloads with more edges
- `edge_count()` — number of unique edges discovered
- `into_hitcount_with_size()` — returns `(buffer, used_size)` so consumers only process meaningful entries
- `hitcount_mut()` — mutable access for external coverage writers (e.g. sancov)
- `into_hitcount()` — preserved for backward compatibility

## Perf

- **Hot path** (known edges): `HashMap::get` is ~same cost as `SipHash + modulo` since both hash the same 3 fields
- **Cold path** (new edges): `HashMap::insert` — ~50ns, happens once per unique edge, amortized over millions of iterations
- **Merge step** (downstream): iterating `[0..used]` instead of full 65536 is a 2-10x speedup for typical edge counts
- **Memory**: HashMap overhead is ~500KB-2MB for 10-50K edges — negligible for fuzzing workloads

## Testing

- 8 new unit tests covering: collision-free IDs, same-edge increment, saturation at 255, capacity exhaustion, reset preserving IDs, `into_hitcount_with_size`, cross-address edge distinction, debug format
- Existing integration test (`test_edge_coverage`) passes unchanged

### DIFF
--- a/src/edge_cov.rs
+++ b/src/edge_cov.rs
@@ -1,9 +1,9 @@
 use alloc::{vec, vec::Vec};
-use alloy_primitives::{map::DefaultHashBuilder, Address, U256};
-use core::{
-    fmt,
-    hash::{BuildHasher, Hash, Hasher},
+use alloy_primitives::{
+    map::{Entry, HashMap},
+    Address, U256,
 };
+use core::fmt;
 use revm::{
     bytecode::opcode::{self},
     interpreter::{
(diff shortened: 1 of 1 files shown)


gakonst:
cc @grandizzy for review

DaniPopes:
moved to foundry-rs/foundry https://github.com/foundry-rs/foundry/blob/c6fcd87df5f50d4595bb2b55e0ac230b88776309/crates/evm/evm/src/inspectors/edge_cov.rs#L23

PR #413, opened 2026-03-18
fix: eip-7708 logs

chihkaiyu:
CallLog discarded the original Log.address during conversion and geth_empty_call_frame() reconstructed it from execution_address(). This is wrong for any log whose emitter differs from the call frame's execution context — most notably EIP-7708 system-level ETH transfer logs, which are emitted from SYSTEM_ADDRESS (0xfffffffffffffffffffffffffffffffffffffffe) rather than the executing contract.

Changes

src/tracing/types.rs
- Add address: Address field to CallLog to preserve the original log emitter
- Update From<Log> to store log.address instead of discarding it
- Fix geth_empty_call_frame() to use log.address instead of self.execution_address()

tests/it/geth.rs
- test_geth_calltracer_logs_eip7708: value-transferring CALL to a precompile under AMSTERDAM — verifies the log address is ETH_TRANSFER_LOG_ADDRESS, not the contract or precompile address
- test_geth_calltracer_logs_address_regular: regular LOG0 emission — verifies the log address is the emitting contract
- test_geth_calltracer_logs_delegatecall: DELEGATECALL regression — verifies the log address is the proxy (execution context), not the implementation (bytecode) address

Breaking

CallLog is a public struct with a new address: Address field. Downstream consumers that construct CallLog manually will need to provide this field.

### DIFF
--- a/src/tracing/types.rs
+++ b/src/tracing/types.rs
@@ -181,6 +181,8 @@ pub struct DecodedCallLog {
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct CallLog {
+    /// The address of the log emitter.
+    pub address: Address,
     /// The raw log data.
     pub raw_log: LogData,
     /// Optional complementary decoded log data.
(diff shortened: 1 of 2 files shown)


mattsse (approved):
I believe this makes sense, we need to mirror the actual log.address for calllogframe

fyi @klkvr @rakita

PR #414, opened 2026-03-23
feat: integrate Inspector frame_start/frame_end hooks

rakita:
## WIP - Do not merge

Integrates `frame_start` and `frame_end` hooks from https://github.com/bluealloy/revm/pull/3518.


mablr (commented):
As https://github.com/bluealloy/revm/pull/3518 was shipped in revm 37.0.0, this should be good now.

PR #422, opened 2026-04-14
fix: check if opcode is valid

decofe:
TracingInspector creates OpCodes via `new_unchecked` at `start_step` to handle unknown/custom opcodes. In `step_end`, `step.op.outputs()` calls `OpCode::info()` which panics with `"opcode not found"` when the opcode byte has no entry in `OPCODE_INFO`.

Guard with `is_valid()` — unknown opcodes default to 0 outputs.

Prompted by: Dragan

### DIFF
--- a/src/tracing/mod.rs
+++ b/src/tracing/mod.rs
@@ -540,9 +540,7 @@ impl TracingInspector {
         if self.config.record_stack_snapshots.is_all()
             || self.config.record_stack_snapshots.is_pushes()
         {
-            // this can potentially underflow if the stack is malformed.
-            // outputs() calls info() which panics on unknown opcodes created via
-            // new_unchecked, so guard with is_valid().
+            // Check if op is valid and return zero
             let outputs = if step.op.is_valid() { step.op.outputs() as usize } else { 0 };
             let start = interp.stack.len().saturating_sub(outputs);
             step.push_stack = Some(interp.stack.data()[start..].into());
--- a/src/tracing/mod.rs
+++ b/src/tracing/mod.rs
@@ -541,9 +541,14 @@ impl TracingInspector {
             || self.config.record_stack_snapshots.is_pushes()
         {
             let outputs = if step.op.is_valid() { step.op.outputs() as usize } else { 0 };
-            let start = interp.stack.len().saturating_sub(outputs);
-            step.push_stack =
-                Some(interp.stack.data().get(start..).unwrap_or_default().into());
+            step.push_stack = Some(
+                interp
+                    .stack
+                    .data()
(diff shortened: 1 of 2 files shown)


src/tracing/mod.rs
```diff
@@ -540,8 +540,11 @@ impl TracingInspector {
         if self.config.record_stack_snapshots.is_all()
             || self.config.record_stack_snapshots.is_pushes()
         {
-            // this can potentially underflow if the stack is malformed
-            let start = interp.stack.len().saturating_sub(step.op.outputs() as usize);
+            // this can potentially underflow if the stack is malformed.
+            // outputs() calls info() which panics on unknown opcodes created via
+            // new_unchecked, so guard with is_valid().
```
rakita:
```suggestion
            // Check if op is valid and return zero
```

src/tracing/types.rs
```diff
@@ -710,7 +710,7 @@ impl CallTraceStep {
             error: self.as_error(),
             gas: self.gas_remaining,
             gas_cost: self.gas_cost,
-            op: self.op.as_str().into(),
```
mattsse, line 713: why is this needed

src/tracing/mod.rs
```diff
@@ -540,8 +540,9 @@ impl TracingInspector {
         if self.config.record_stack_snapshots.is_all()
             || self.config.record_stack_snapshots.is_pushes()
         {
-            // this can potentially underflow if the stack is malformed
-            let start = interp.stack.len().saturating_sub(step.op.outputs() as usize);
+            // Check if op is valid and return zero
+            let outputs = if step.op.is_valid() { step.op.outputs() as usize } else { 0 };
+            let start = interp.stack.len().saturating_sub(outputs);
             step.push_stack = Some(interp.stack.data()[start..].into());
```
mattsse: this does not return 0

why is this the right fix and not using get() -> Option

src/tracing/types.rs
rakita: as_str uses info in the background, so it can panic, but to_string is not a way to fix it. Will make change here and use op.is_valid()

src/tracing/mod.rs
rakita: Like 544 is a fix, as step.op.outputs() can panic as OpCode is created as `unsafe` 

But this can be simplified with get(

rakita: @mattsse it is slightly better now, pls take a look.

DaniPopes:
https://github.com/bluealloy/revm/pull/3566 this is probably better

mattsse (approved):

PR #438, opened 2026-05-15
fix: construct JS tracer bigints directly

mattsse:
## Summary

Construct internal JS tracer BigInt values directly with Boa `JsBigInt` instead of reading and calling the mutable global `bigint` alias.

The public `bigint` alias remains registered for tracer compatibility, but internal Rust bindings no longer depend on tracer-controlled global state.

## Root Cause

`to_bigint` looked up `globalThis.bigint` and called it for internal conversions. A custom tracer could overwrite or poison that property, causing stack, contract, context, and DB BigInt bindings to fail unexpectedly.

## Impact

This keeps tracer-visible compatibility while making internal BigInt conversion deterministic and isolated from user JavaScript mutations.


src/tracing/js/builtins.rs
```diff
@@ -194,11 +195,13 @@ pub(crate) fn bytes_to_fb<const N: usize>(mut bytes: &[u8]) -> FixedBytes<N> {
     FixedBytes::left_padding_from(bytes)
 }
 
-/// Converts a U256 to a bigint using the global bigint alias.
-pub(crate) fn to_bigint(value: U256, ctx: &mut Context) -> JsResult<JsValue> {
-    let bigint = ctx.global_object().get(js_string!("bigint"), ctx)?;
-    let Some(bigint) = bigint.as_callable() else { return Ok(JsValue::undefined()) };
-    bigint.call(&JsValue::undefined(), &[JsValue::from(js_string!(value.to_string()))], ctx)
+/// Converts a U256 to a Boa bigint value.
+pub(crate) fn to_bigint(value: U256, _ctx: &mut Context) -> JsResult<JsValue> {
```
DaniPopes: can just drop the arg

PR #441, opened 2026-05-16
fix: validate JS tracer memory indexes

mattsse:
## Summary

Validate JS tracer memory indexes before casting them to `usize` or passing them into memory slicing.

`log.memory.slice(...)` and `log.memory.getUint(...)` now reject non-finite, negative, too-large, and out-of-bounds numeric inputs before reading from the memory snapshot.

## Root Cause

The memory binding converted JavaScript numeric values to Rust indexes before fully rejecting invalid values. Inputs like `Infinity` and `NaN` could pass the floating-point checks and become invalid `usize` ranges.

## Impact

Invalid tracer memory inputs now fail as JavaScript errors instead of reaching unchecked memory access paths in the host binding.


DaniPopes (commented):
aren't there methods on boa to convert?

DaniPopes (approved):

PR #449, opened 2026-05-28
Optimize JS tracer callback object reuse

haardikk21:
## TL;DR

- The old `js-tracer` hot path rebuilt fresh Boa wrapper objects on every `step`, `fault`, `enter`, and `exit` callback.
- That per-callback object construction was the main source of overhead, especially for opcode-heavy tracers.
- The fix here is to create stable reusable JS objects once and update shared Rust-side backing state before each callback.
- The external JS tracer API does not change; scripts still use the same Geth-style methods and object shapes.
- This PR also adds benchmark coverage for synthetic accessor-heavy workloads, Rundler-shaped workloads, real Rundler tracer scripts, and a real mainnet AA transaction replay.
- The optimization materially improves all measured cases:
- `step_noop`: `157.18 ms` -> `4.1913 ms` (`37.5x`)
- `step_accessors`: `230.39 ms` -> `38.457 ms` (`6.0x`)
- `rundler_v06_real`: `41.989 ms` -> `15.463 ms` (`2.72x`)
- `mainnet_aa_v06_real`: `9.7643 ms` -> `3.6274 ms` (`2.69x`)
- The mainnet AA improvement is smaller than the synthetic upper-bound cases because that benchmark includes more non-tracer work, not because the benchmark is broken.

`cargo bench --bench js_tracer --features js-tracer` to validate

Full writeup: https://hackmd.io/@vFMCdNzHQNqyPq_Rej0IIw/B16xdMPeGe


src/tracing/js/mod.rs
```diff
@@ -1104,11 +1005,9 @@ mod tests {
         }"#;
         let res = run_trace(code, None, true);
 
-        // The bytecode is: PUSH1 0x01, PUSH1 0x01, STOP
-        // Expected costs: PUSH1=3, PUSH1=3, STOP=0
```
DaniPopes, line 1108: these removals look like mishandled merge conflicts, please fix

haardikk21: sorry about that, fixing

src/tracing/js/bindings.rs
```diff
@@ -678,6 +1404,7 @@ pub(crate) struct CallFrame {
 }
 
 impl CallFrame {
+    #[cfg(test)]
     pub(crate) fn into_js_object(self, ctx: &mut Context) -> JsResult<JsObject> {
```
DaniPopes: can you please also update tests to use the new stuff so we dont have two copies of the js object creation?

haardikk21: Yeah, left it there to run before/after for the bench. Will update in the AM to remove it

haardikk21: updated

DaniPopes (approved):
lgtm, needs CI

decofe:
I tried to validate the benchmark numbers locally, but I could not reproduce the full set from the PR as-is.

A couple of things seem to be missing from the repro instructions:

- The benchmark only runs with `--features js-tracer`; `cargo bench --bench js_tracer` fails without it.
- `rundler_v06_real` and `mainnet_aa_v06_real` are gated on `RUNDLER_V06_TRACER_PATH`, but I do not see the Rundler tracer script or setup steps in the PR. Without that env var, those cases are silently skipped.

Can you add the exact command/env setup you used, and either commit the tracer fixture or document where to get the exact script? The synthetic cases do show the expected direction locally, but the real benchmark claims are not reproducible from the PR alone right now.

haardikk21:
> I tried to validate the benchmark numbers locally, but I could not reproduce the full set from the PR as-is.
> 
> A couple of things seem to be missing from the repro instructions:
> 
> * The benchmark only runs with `--features js-tracer`; `cargo bench --bench js_tracer` fails without it.
> * `rundler_v06_real` and `mainnet_aa_v06_real` are gated on `RUNDLER_V06_TRACER_PATH`, but I do not see the Rundler tracer script or setup steps in the PR. Without that env var, those cases are silently skipped.
> 
> Can you add the exact command/env setup you used, and either commit the tracer fixture or document where to get the exact script? The synthetic cases do show the expected direction locally, but the real benchmark claims are not reproducible from the PR alone right now.

Yeah the JS tracer feature must be enabled for this bench to run

My bad on the rundler path not being specified. You can run the following to run the bench on the new version:

```sh
# Clone and pin Rundler
git clone https://github.com/alchemyplatform/rundler.git /tmp/rundler
git -C /tmp/rundler checkout 073b093112e8b27dbf62ef6ede7664526e09243b # If you want to match exactly the version I used, though not necessary

# Build the tracer JS files
cd /tmp/rundler/crates/sim/tracer
npm install
npm run build

# This should produce:
# /tmp/rundler/crates/sim/tracer/dist/validationTracerV0_6.js
# /tmp/rundler/crates/sim/tracer/dist/validationTracerV0_7.js

# Back in revm-inspectors
cd /path/to/revm-inspectors

RUNDLER_V06_TRACER_PATH=/tmp/rundler/crates/sim/tracer/dist/validationTracerV0_6.js \
RUNDLER_V07_TRACER_PATH=/tmp/rundler/crates/sim/tracer/dist/validationTracerV0_7.js \
cargo bench --bench js_tracer --features js-tracer -- --noplot
```

We can either:
- vendor the JS scripts into this repo directly
- add Makefile/Justfile step to clone the repo, build, and vendor the files locally with them being in gitignore
- remove the rundler benches entirely

DaniPopes (commented):
geth does the same, and documents that objects shouldn't be relied upon directly (https://geth.ethereum.org/docs/developers/evm-tracing/custom-tracer) so this works, but i think we just need a regression test to verify this behavior, as in the function always receives the same object

DaniPopes:
> We can either:
> 
>     * vendor the JS scripts into this repo directly
> 
>     * add Makefile/Justfile step to clone the repo, build, and vendor the files locally with them being in gitignore
> 
>     * remove the rundler benches entirely

let's vendor them in the bench dir and use include_str!

haardikk21:
Vendored, don't need the paths anymore. Just run `cargo bench --bench js_tracer --features js-tracer`

Also added a simple regression test to ensure the same JS object for `log` and `db` is received each time per step

DaniPopes (approved):

mattsse (approved):
lgtm

PR #453, opened 2026-06-02
fix: remove transfer inspector reverted logs

Soubhik-10:
Updates `TransferInspector` so synthetic transfer-log mode cleans up frame-local transfers/logs when `CALL` or `CREATE` frames revert or halt.

This keeps `TransferInspector::with_logs(true)` aligned with EVM revert semantics: transfer logs inserted for failed frames are removed instead of leaking into the final execution result.

## Details

- Adds frame checkpoints to `TransferInspector` when synthetic log insertion is enabled.
- On `call_end` / `create_end`, successful outcomes keep collected transfers/logs.
- Reverted or halted outcomes roll back:
  - collected transfer operations
  - synthetic logs inserted into the journal
- Adds regression coverage for reverted `CALL` and reverted `CREATE`.

## Why

`eth_simulateV1 traceTransfers` represents ETH transfers as synthetic ERC20-style logs. These logs are RPC artifacts, not contract-emitted logs, so failed internal frames should not leak them into successful parent results.

### DIFF
--- a/src/transfer.rs
+++ b/src/transfer.rs
@@ -24,7 +24,7 @@ pub const TRANSFER_EVENT_TOPIC: B256 =
 pub struct TransferInspector {
     internal_only: bool,
     transfers: Vec<TransferOperation>,
-    call_checkpoints: Vec<TransferCheckpoint>,
+    checkpoints: Vec<TransferCheckpoint>,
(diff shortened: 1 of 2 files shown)


mattsse (approved):
LGTM

PR #466, opened 2026-07-01
feat(writer): render named-but-undecoded logs with raw topics/data

0xMars42:
## Motivation

When an event is emitted with data that is not ABI-encoded, its `topic0` can still match a known event while `decode_log` fails. Foundry's decoder then produces `DecodedCallLog { name: Some(event_name), params: None }` (foundry-rs/foundry#15087, for foundry-rs/foundry#10451).

`TraceWriter::write_log` branched only on `decoded_name()`, so that case rendered `emit Exchange()` and never reached the raw topics/data branch: the raw data was dropped. On its own the decoder change trades "raw data, no name" for "name, no data".

## Change

Match on `(decoded_name(), decoded_params())`:
- `(Some, Some)`: decoded render, unchanged
- `(Some, None)`: print the event name, then the raw topics/data (new)
- `(None, _)`: raw render, unchanged

The raw topics/data block moves into a small `write_raw_log` helper shared by the named-but-undecoded and fully-undecoded paths. Rendering follows the named header plus raw topics/data layout.

## Notes

Existing writer snapshots exercise `(Some, Some)` and `(None, _)`; the Counter events carry indexed topics, so `params` is always set and the new arm is not hit by them. `cargo test --all-features` passes unchanged. Happy to add a dedicated snapshot for the `(Some, None)` case if you would prefer one in this PR.

### DIFF
--- a/src/tracing/writer.rs
+++ b/src/tracing/writer.rs
@@ -322,37 +322,56 @@ impl<W: Write> TraceWriter<W> {
         let log_style = self.log_style();
         self.write_branch()?;
 
-        if let Some(name) = log.decoded_name() {
-            write!(self.writer, "emit {name}({log_style}")?;
-            if let Some(params) = log.decoded_params() {
+        match (log.decoded_name(), log.decoded_params()) {
+            // Fully decoded event: `emit Name(param: value, ...)`.
+            (Some(name), Some(params)) => {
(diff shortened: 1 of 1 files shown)


0xMars42:
Duplicate of #469, which also adds a dedicated test for the named-but-undecoded case. Closing this one in favor of that PR.

PR #467, opened 2026-07-02
fix(geth): derive log positions from trace ordering

### LINKED ISSUE #187: A better approach for populating `log.position` in Geth trace.
#186 added it directly to `CallLog`, but `CallTraceNode` already has `ordering` which should be enough to populate such parameters.

#184 has a ready approach for it, but it is recursive and we would like to avoid recursion here due to SO issues

Osraka:
Closes #187.

## Summary

- Derive geth `CallLogFrame.position` from `CallTraceNode::ordering` when building call frames.
- Stop writing the position into `CallLog` during tracing.
- Keep the existing `CallLog::position` field as a compatibility fallback instead of removing public API.
- Add a regression test for interleaved log/call ordering.

## Rationale

`CallLogFrame.position` is relative to child calls in the same frame. The trace node already records the order of calls and logs, so the geth output can compute the position at serialization time without adding recursion to the call-frame builder.

The patch intentionally leaves the existing non-recursive call-frame rollup unchanged.

## Tests

- `cargo test test_geth_calltracer_log_positions_are_derived_from_ordering`
- `cargo test --test it geth::`
- `cargo test`
- `cargo fmt --check` (passes; stable rustfmt prints warnings for nightly-only config options)

### DIFF
--- a/src/tracing/mod.rs
+++ b/src/tracing/mod.rs
@@ -646,11 +646,7 @@ where
             let log_count = self.log_count();
             let trace = self.last_trace();
             trace.ordering.push(TraceMemberOrder::Log(trace.logs.len()));
-            trace.logs.push(
-                CallLog::from(log)
-                    .with_position(trace.children.len() as u64)
-                    .with_index(log_count as u64),
-            );
+            trace.logs.push(CallLog::from(log).with_index(log_count as u64));
         }
     }
 
--- a/src/tracing/types.rs
+++ b/src/tracing/types.rs
@@ -189,7 +189,7 @@ pub struct CallLog {
     pub decoded: Option<Box<DecodedCallLog>>,
     /// The position of the log relative to subcalls within the same trace.
     pub position: u64,
-    /// The position of the log relative to subcalls within the same trace.
+    /// The index of the log in the transaction trace.
     pub index: u64,
 }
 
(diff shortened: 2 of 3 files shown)


Osraka:
Marked this ready for review after rechecking it against current main.\n\nValidation run locally:\n- `cargo test test_geth_calltracer_log_positions_are_derived_from_ordering`\n- `cargo test test_geth_calltracer_logs`\n- `cargo test --test it geth::`\n- `cargo test --lib`\n- `cargo test`\n- `cargo fmt --check`\n- `git diff --check origin/main...HEAD`\n\nThe branch is one commit ahead of `origin/main` and not behind it. The implementation keeps the call-frame rollup non-recursive and leaves `CallLog::position` in place as a compatibility fallback.

Osraka:
Refreshed the branch onto current `main` (revm 43 / alloy 2.4) and preserved the upstream EIP-8037 tracing tests while resolving the integration-test import conflict.

Validation on the rebased commit:
- targeted log-position regression: 1 passed
- Geth integration tests: 17 passed
- `cargo test --workspace --all-features`: 64 unit + 39 integration tests passed
- `cargo clippy --workspace --all-targets --all-features`
- `cargo fmt --check`
- `git diff --check origin/main...HEAD`

mattsse (commented):
can you elaborate on what this fixes

need an example where this deviates from other clients

Osraka:
@mattsse You're right to ask. I checked this against Geth's native call tracer and I can't produce a client-output deviation from current main.

Geth assigns `Position = len(currentFrame.Calls)` in `OnLog`: https://github.com/ethereum/go-ethereum/blob/bbb9119caaefbbf619363493e54cc8362d65188d/eth/tracers/native/call.go#L237-L256

Current revm-inspectors records `trace.children.len()` when the log is captured. `CallTraceArena::push_trace` appends the `TraceMemberOrder::Call` entry and child together, so for inspector-produced traces that value is equivalent to scanning `ordering` and counting preceding calls. The `position: 99` regression test only demonstrates behavior for a manually inconsistent `CallTraceNode`; the normal tracing path does not produce that state.

So this is better described as the non-recursive state deduplication contemplated by #187, not a behavioral fix or an interoperability correction. On further review, stopping population of the public `CallLog::position` field could also change direct `CallTraceNode` consumers even though the serialized Geth frame remains unchanged.

I don't want to overstate the patch, so I'm moving it back to draft. If the #187 cleanup is still desired, I can rework it once we agree on how the public `CallLog::position` field should be handled; otherwise I'll close it.

PR #475, opened 2026-08-04
feat(tracing): support EIP-8037 trace fields

decofe:
Implements the EIP-8037 tracing additions from [ethereum/execution-apis#852](https://github.com/ethereum/execution-apis/pull/852), using [Alloy #4113](https://github.com/alloy-rs/alloy/pull/4113) as a temporary dependency patch.

Changes:
- expose regular/state gas accounting on default and call traces;
- emit per-opcode `stateGasCost` and `stateGasReservoir` values;
- support the `stateGasTracer` built-in tracer;
- add an integration test for the new tracer.

The Alloy patch is intentionally pinned by commit and can be removed once #4113 is merged. The additive RPC fields preserve backwards-compatible decoding for existing callers.

Tests: `cargo test`
Clippy: `cargo +nightly clippy --all-targets --all-features -- -D warnings`

Prompted by: @mattsse

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,3 +72,7 @@ std = [
 ]
 serde = ["dep:serde", "revm/serde"]
 js-tracer = ["dep:boa_engine", "dep:boa_gc"]
+
+[patch.crates-io]
+# Temporary dependency patch for the pending EIP-8037 RPC types.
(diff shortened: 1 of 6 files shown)


onbjerg:
merged in #476

PR #479, opened 2026-08-11
fix: normalize access list collection

Pana:
Normalizes seeded access lists with alloy-eip2930's native helpers and removes entries already warmed by execution. This keeps output deterministic and prevents excluded addresses supplied in the seed from leaking into the result.

### DIFF
--- a/src/access_list.rs
+++ b/src/access_list.rs
@@ -1,11 +1,11 @@
 use alloc::collections::BTreeSet;
+use alloy_eip2930::{AccessList, AccessListItem};
 use alloy_primitives::{
     map::{HashMap, HashSet},
     Address, TxKind, B256,
 };
 use revm::context::transaction::AuthorizationTr;
 
-use alloy_rpc_types_eth::{AccessList, AccessListItem};
 use revm::{
     bytecode::opcode,
     context::JournalTr,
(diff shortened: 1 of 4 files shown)


Pana:
To align with the behavior of geth

src/access_list.rs
```diff
@@ -107,6 +111,7 @@ impl AccessListInspector {
         let auth_addrs = context.tx().authorization_list().flat_map(|a| a.authority());
 
         self.excluded = [from, to].into_iter().chain(precompiles).chain(auth_addrs).collect();
+        self.touched_slots.retain(|address, _| !self.excluded.contains(address));
```
mattsse, line 114: why do we need this?

src/access_list.rs
```diff
@@ -40,14 +40,12 @@ impl AccessListInspector {
...
-                .0
-                .into_iter()
-                .map(|v| (v.address, v.storage_keys.into_iter().collect()))
-                .collect(),
+        let mut touched_slots: HashMap<Address, BTreeSet<B256>> = Default::default();
+        for item in access_list.0 {
+            touched_slots.entry(item.address).or_default().extend(item.storage_keys);
         }
+
+        Self { excluded: Default::default(), touched_slots }
```
mattsse, line 48: this is so that the AL becomes unique?

Pana: User input may contain duplicate addresses; this operation is mainly for deduplication and merging.

src/access_list.rs
Pana: The collected results should exclude the addresses in "excluded".

Pana: The initial provided AL should also excluded by "excluded"

DaniPopes (approved):

mattsse (approved):

PR #493, opened 2026-09-05
perf(js): avoid per-step allocations and copies

mattsse:
Stacked on #492.

Every step cloned the interpreter stack, which allocates the full 1024 slot capacity (32KB) since `Stack::clone` goes through `Stack::new`, and allocated six more objects to hand the stack, memory, state and database to the JS callbacks. In the profile of a no-op `step` callback, malloc and dropping these objects were about half of the time.

The reusable JS objects now keep persistent slots that point at the interpreter only for the duration of a callback. Instead of copying the stack and memory before the opcode executes, `step` saves only the parts the opcode can overwrite, its stack inputs (per the opcode table, at most 17 items) and the memory range it writes, and `step_end` reconstructs the pre-execution view geth exposes to `step` from the post-execution state. This holds for halted opcodes too, since they pop at most their inputs and never write partially, and for calls, whose return data is only written once the child frame returns.

Borrowed stack, memory, state, and database access is restricted to scoped callbacks. The input borrows remain live for the entire callback, and private cleanup guards revoke access on return or unwinding; callers cannot retain or forget a guard. The branch also includes #492’s callback-wrapper reset so JS object properties do not leak between transactions.

BigInts are built from limbs instead of a decimal string round trip, `Uint8Array` arguments are read from their buffer instead of one element at a time, and opcode and call kind names are cached JS strings. The opcode-name table is initialized on the first `log.op.toString()` lookup and shared across `fuse` resets, so tracers that never request opcode names avoid building it.

A step with an empty callback went from 7 allocations (33KB) to none. Benchmarks against main (`mainnet_aa_v06_real` creates a new inspector per iteration, so script parsing dominates it):

| bench | change |
|---|---|
| step_noop | -73% |
| step_accessors | -34% |
| rundler_v06_style | -48% |
| rundler_v07_style | -60% |
| rundler_v06_real | -45% |
| rundler_v07_real | -58% |
| mainnet_aa_v06_real | -18% |

Mirrored to https://github.com/alloy-rs/evm2/pull/411.

### DIFF
--- a/src/tracing/js/builtins.rs
+++ b/src/tracing/js/builtins.rs
@@ -1,11 +1,15 @@
 //! Builtin functions
 
-use alloc::{borrow::Cow, format, string::ToString, vec::Vec};
+use crate::tracing::types::CallKind;
+use alloc::{borrow::Cow, format, vec::Vec};
 use alloy_primitives::{hex, map::HashSet, Address, FixedBytes, B256, U256};
 use boa_engine::{
-    builtins::{array_buffer::ArrayBuffer, typed_array::TypedArray},
+    builtins::{
+        array_buffer::ArrayBuffer,
+        typed_array::{TypedArray, TypedArrayKind},
+    },
     js_string,
-    object::builtins::{JsArray, JsArrayBuffer, JsTypedArray, JsUint8Array},
+    object::builtins::{AlignedVec, JsArray, JsArrayBuffer, JsTypedArray, JsUint8Array},
     property::Attribute,
     Context, JsArgs, JsBigInt, JsError, JsNativeError, JsResult, JsString, JsValue, NativeFunction,
     Source,
@@ -120,6 +124,11 @@ pub(crate) fn bytes_from_value(val: JsValue, context: &mut Context) -> JsResult<
     if let Some(obj) = val.as_object() {
         if obj.is::<TypedArray>() {
             let array: JsTypedArray = JsTypedArray::from_object(obj)?;
+            if array.kind() == Some(TypedArrayKind::Uint8) {
+                if let Some(bytes) = uint8_array_bytes(&array, context)? {
+                    return Ok(bytes);
+                }
+            }
             let len = array.length(context)?;
             let mut buf = Vec::with_capacity(len);
             for i in 0..len {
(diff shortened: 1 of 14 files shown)


mattsse:
[P2] Tie scope guards to the borrowed data’s lifetime (`src/tracing/js/bindings.rs:198`).

`ReusableStepLog::enter()` returns `StepScope` tied only to `&self`, while the stack and memory borrows have independent lifetimes and are internally erased to static references. A compile-only reproduction confirms that safe Rust can drop the stack and memory allocations before the guard; subsequent callback access, or dropping the guard’s memory `Ref`, can then use freed storage. `ReusableEvmDb::enter()` has the same lifetime gap for state and database borrows.

The current production callers keep those inputs alive, but the safe APIs do not enforce that requirement. Bind the input borrows to the scope, and preferably keep guards internal to scoped callback helpers so callers cannot forget them. Add coverage for borrow revocation after normal returns and unwinding, alongside compile-time lifetime checks.

All 109 tests passed, and temporary full-snapshot comparisons passed across 768 generated opcode cases. This does not exercise the dangling access, which was deliberately not executed.

src/tracing/js/builtins.rs
```diff
@@ -214,12 +274,22 @@ pub(crate) fn bytes_to_fb<const N: usize>(mut bytes: &[u8]) -> FixedBytes<N> {
...
+    let limbs = value.as_limbs();
+    // Most values fit into a machine integer, for which `num-bigint` has cheap constructors.
+    // Larger values (e.g. addresses and hashes) are constructed from their big endian bytes,
+    // avoiding a decimal round trip through strings.
+    let big = if limbs[1] | limbs[2] | limbs[3] == 0 {
+        JsBigInt::from(limbs[0])
+    } else if limbs[2] | limbs[3] == 0 {
+        JsBigInt::from(((limbs[1] as u128) << 64) | limbs[0] as u128)
+    } else {
+        JsBigInt::from(num_bigint::BigInt::from_bytes_be(
```
DaniPopes: we could use this https://docs.rs/ruint/latest/src/ruint/support/num_bigint.rs.html

src/tracing/js/builtins.rs
```diff
@@ -214,12 +274,22 @@ pub(crate) fn bytes_to_fb<const N: usize>(mut bytes: &[u8]) -> FixedBytes<N> {
...
-        JsError::from_native(
-            JsNativeError::error().with_message("failed to convert U256 to BigInt"),
-        )
-    })
+pub(crate) fn to_bigint(value: U256) -> JsValue {
+    let limbs = value.as_limbs();
+    // Most values fit into a machine integer, for which `num-bigint` has cheap constructors.
+    // Larger values (e.g. addresses and hashes) are constructed from their big endian bytes,
+    // avoiding a decimal round trip through strings.
+    let big = if limbs[1] | limbs[2] | limbs[3] == 0 {
```
DaniPopes: you should do u64::try_from, u128::try_from instead of manually checking limbs

src/tracing/js/builtins.rs
mattsse: Updated in 07d347c: the large-value path now uses ruint’s `From<U256> for num_bigint::BigInt`, with num-bigint support enabled only for js-tracer. Added a mixed-limb conversion case to catch byte-order mistakes. Mirrored in alloy-rs/evm2#411.

src/tracing/js/builtins.rs
mattsse: Updated in 07d347c to use `u64::try_from(value)` and `u128::try_from(value)`, retaining the small-integer constructors without manually checking or combining limbs. The conversion test covers u64::MAX, u64::MAX + 1, u128::MAX, u128::MAX + 1, and U256::MAX. Mirrored in alloy-rs/evm2#411.

DaniPopes (approved):

PR #494, opened 2026-09-05
perf(geth): propagate log visibility through call frames

mattsse:
With `withLog`, callTracer and ERC-7562 walked from every call back to the root to check whether an ancestor reverted. A successful chain therefore did quadratic work even after #489's linear call-frame rollup.

Carry log visibility in a named `PendingCallFrame` field alongside each frame and inherit the parent's flag during the existing forward pass. The flag replaces the redundant arena index, so this adds no allocation. Root-frame behavior and child/log ordering are preserved.

Criterion on an Apple M5 Max, rustc 1.99.0-nightly (3d50c25bc), against #489 (`23477fa`). These measure frame construction and destruction from prebuilt successful call chains, with one log per call; execution and recording are excluded.

| Builder / nested calls | Before | After | Reduction |
|---|---:|---:|---:|
| callTracer / 8 | 0.820 µs | 0.763 µs | 7% |
| callTracer / 64 | 10.62 µs | 6.55 µs | 38% |
| callTracer / 256 | 88.57 µs | 27.74 µs | 69% |
| callTracer / 1024 | 1.503 ms | 0.113 ms | 92% |
| ERC-7562 / 64 | 11.56 µs | 7.47 µs | 35% |
| ERC-7562 / 1024 | 1.520 ms | 0.123 ms | 92% |

Logging-disabled callTracer runs were essentially unchanged. The deepest cases expose the scaling issue rather than represent typical transactions.

Adds regression coverage for reverted and halted branches, successful descendants of failed ancestors, log-free intermediate calls, later successful siblings, root failure, disabled logs, and child/log ordering in both builders.

Mirrored in https://github.com/alloy-rs/evm2/pull/408.


benches/geth_log_ancestry.rs
```diff
@@ -0,0 +1,71 @@
+//! Isolates call-frame construction from execution and trace recording.
+
+use alloy_primitives::{Address, Log};
```
DaniPopes: remove all benchmarks

src/tracing/builder/geth.rs
```diff
@@ -233,12 +233,15 @@ impl<'a> GethTraceBuilder<'a> {
         // traces are identified by their index in the arena
         // so we can populate the call frame tree by walking up the call tree
         let mut call_frames = Vec::with_capacity(self.nodes.len());
-        call_frames.push((0, root_call_frame));
+        call_frames.push((include_logs && !main_trace_node.trace.is_error(), root_call_frame));
```
DaniPopes: what's the point of storing this bool? i dont see it being used anymore, below too

DaniPopes (changes requested):

src/tracing/builder/geth.rs
mattsse: The flag is read in the forward pass through `call_frames[parent].logs_visible` (and `frames[parent].logs_visible` for ERC-7562), so each child inherits whether an ancestor failed without walking back to the root. It is no longer needed during the reverse tree assembly. I replaced the tuple with `PendingCallFrame { logs_visible, frame }` in 58dc7e0 to make those reads explicit and documented why visibility is independent of whether the frame contains logs. The existing regression test covers a failed ancestor through a log-free intermediate call, later successful siblings, and root failure in both builders.

DaniPopes (approved):

PR #495, opened 2026-09-05
perf(geth): stream opcode steps during trace construction

mattsse:
The geth opcode trace builder copied every step reference into a temporary queue, then scanned the inserted range again to assign child calls. A flat 20,000-step trace needed a 640 KB queue before building its output.

Walk each call's steps directly and suspend only the parent iterator while visiting a child. Scratch space now scales with active call depth, with no traversal allocation for a flat trace. This also removes the separate pass for matching call opcodes to children.

Criterion on an Apple M5 Max, rustc 1.99.0-nightly (3d50c25bc), against main (`9c3118c`), including the opcode-limit fix from #487. These measure opcode-frame construction and destruction from prebuilt traces, excluding execution, recording, and JSON serialization. Stack and memory snapshots are absent in these fixtures.

| Trace shape | Storage capture | Before | After | Reduction |
|---|---|---:|---:|---:|
| 20,000 flat steps | enabled | 420.0 µs | 406.0 µs | 3% |
| 20,000 flat steps | disabled | 260.1 µs | 221.8 µs | 15% |
| 2,000 sibling calls × 20 steps | enabled | 941.3 µs | 896.7 µs | 5% |
| 2,000 sibling calls × 20 steps | disabled | 579.3 µs | 509.9 µs | 12% |
| 64 nested calls × 128 steps | enabled | 176.6 µs | 166.1 µs | 6% |
| 64 nested calls × 128 steps | disabled | 108.7 µs | 93.1 µs | 14% |

Adds regression coverage for parent resumption, all six call/create opcodes, an empty precompile, empty creation, a reverting grandchild, a call that fails before recording a child, and empty traces. It checks exact step/depth ordering, stack and return data, gas, and storage snapshots across delegate calls. Limits at several nested-call boundaries must produce the exact prefix of the unlimited trace; zero remains unlimited.

Mirrored in https://github.com/alloy-rs/evm2/pull/405.


benches/geth_steps.rs
```diff
@@ -0,0 +1,96 @@
+//! Geth opcode-frame construction, excluding execution and recording.
+
+use alloy_primitives::Bytes;
```
DaniPopes: remove benchmarks

DaniPopes (approved):

PR #496, opened 2026-09-05
perf(geth): skip storage lookups for unrelated opcodes

mattsse:
With storage capture enabled (the default), the geth opcode trace builder looked up the execution address in its storage map for every opcode. Most steps neither update storage nor emit a storage snapshot, so this also created unused entries for contracts without storage access.

Only access the cache when the step has a recorded storage change or is SLOAD/SSTORE. SLOAD/SSTORE still emit an empty snapshot when no change was recorded, and custom storage-change metadata on other opcodes is still applied.

Criterion on an Apple M5 Max, rustc 1.99.0-nightly (3d50c25bc), measured on top of #495 (`8db6c02`), including the opcode-limit fix from #487. These measure opcode-frame construction and destruction from prebuilt traces, excluding execution, recording, and serialization. Stack and memory snapshots are absent.

| Trace shape, storage capture enabled | Before | After | Reduction |
|---|---:|---:|---:|
| 20,000 flat steps | 406.0 µs | 225.9 µs | 44% |
| 20,000 steps, one SLOAD every 16 steps | 434.6 µs | 271.5 µs | 38% |
| 2,000 sibling calls × 20 steps | 896.7 µs | 513.5 µs | 43% |
| 64 nested calls × 128 steps | 166.1 µs | 93.5 µs | 44% |

The mixed fixture repeatedly accesses one slot. The other fixtures have no storage accesses and show the avoidable lookup overhead. Storage-disabled cases were about 0.6–1.0% slower in this run (for example, the flat case moved from 221.8 to 224.1 µs).

Adds regression coverage for empty reads, custom change metadata, and disabled storage capture. The nested-call ordering and delegate-call storage tests from #495 also pass.

Mirrored in https://github.com/alloy-rs/evm2/pull/406.


benches/geth_steps.rs
```diff
@@ -1,14 +1,14 @@
 //! Geth opcode-frame construction, excluding execution and recording.
 
-use alloy_primitives::Bytes;
```
DaniPopes: remove benchmarks

DaniPopes (approved):

PR #505, opened 2026-09-14
perf(tracing): compact step recording when snapshots are disabled

mattsse:
Moves opt-in memory, returndata and immediate-byte snapshots into a per-call side table, reducing `CallTraceStep` from 232 to 128 bytes. Buffer indices stay aligned when capture settings change mid-frame, and `fuse()` reuses both step and buffer vectors. Also skips journal-length reads when state diffs are disabled.

Local `struct_logger` spot-check: 103,501 steps over 301 frames, second of two alternating base/branch runs (20 samples, 0.5s warmup, 1s measurement; JSON serialization excluded):

| Recording | Base | Branch | Change |
| --- | ---: | ---: | ---: |
| Fresh, stack/storage disabled | 2.660 ms | 1.827 ms | -31.3% |
| Fresh, default | 3.725 ms | 3.037 ms | -18.5% |
| Fresh, memory/returndata enabled | 3.965 ms | 4.115 ms | +3.8% |
| Reused, stack/storage disabled | 1.937 ms | 1.433 ms | -26.0% |
| Reused, default | 2.959 ms | 2.618 ms | -11.5% |
| Reused, memory/returndata enabled | 3.292 ms | 3.267 ms | -0.8% |

Breaking Rust API and serialized trace schema: `CallTraceStep::{memory, returndata, immediate_bytes}` move to `CallTrace::step_buffers` / `buffers_at(step_idx)`, and `state_gas_spent` becomes inspector scratch state. A missing `step_buffers` field defaults to empty for older traces without snapshots; older inline snapshots require migration. Geth struct-log and vmTrace output is unchanged.


mattsse:
End-to-end measurement with reth's RPC replay benchmark (`debug_traceBlockByNumber` on 20 mainnet blocks, one `callTracer` and one struct-logger record per block with `disableStack` and `disableStorage`, 2 rps, 2 run pairs, parity 80/80). Baseline pins revm-inspectors `main` (8865141), feature pins this branch (bad1891e), so only this change is measured. Run: https://github.com/paradigmxyz/reth/actions/runs/34795784560

| Metric | main | this PR | Change |
|---|---|---|---|
| structlog mean | 862.72ms | 834.00ms | -3.33% (±0.42%, A/A +0.13%) |
| callTracer mean (control) | 77.63ms | 76.34ms | -1.66% (inside the 2.5% floor) |
| CPU / request, both | 320.40ms | 299.39ms | -6.56% (±0.30%, A/A -0.12%) |
| P50, both | 203.74ms | 192.20ms | -5.66% |

Struct-logger responses are ~119 MB each, so the remaining time is dominated by serialization and the response buffer; the control row does not move, as expected.

PR #513, opened 2026-09-22
chore(ci): protect dependency installs with Aegis

decofe:
Add pinned Aegis/runner protection before dependency installs in every CI job that builds, checks or installs Rust tooling, with job-scoped OIDC permissions. Follow the Alloy patterns for checksum-verified cargo-hack and host-side cargo-deny while preserving the existing matrix and lockfile policy.

Fork PRs without OIDC retain the shared action's skip behavior; those runs are not fully Aegis-enforced.

Prompted by: @grandizzy


mablr (commented):
sgtm

PR #518, opened 2026-09-23
fix(tracing): cap call input bytes to prevent tracer OOM

christophercampbell:
Without a bound, the tracer copies every CALL's calldata into memory, and calldata size is nearly free in gas. We observed in production a crafted transaction that OOMs a node serving debug_traceTransaction, so this is an exploitable vector.

This PR adds an opt-in per-frame cap and a total budget on the call input copied into the trace arena. Both default to unlimited, so behavior is unchanged unless a caller opts in.

## Testing

New integration suite driving real EVM bytecode: a 64 KiB CALL arriving as a zero-copy CallInput::SharedBuffer, a loop of five 1 KiB CALLs that only the total budget can bound, owned transaction calldata, and CREATE init code. Covers the unlimited default, per-frame truncation reporting the true length, budget exhaustion leaving the call tree complete, reset on fuse(), and merge semantics.

```
cargo test --workspace --all-features
cargo clippy --workspace --all-targets --all-features
cargo hack check --feature-powerset --depth 2
```


mattsse:
@christophercampbell thanks for this

this makes sense, but I'll implement this slightly differently.

do you know which exact tracer was used here? regular calltracer?

christophercampbell:
@mattsse Yes, the regular callTracer via debug_traceTransaction

banteg:
The 64 KiB single-CALL test is a useful reproduction, but the gas-bounded worst case for this copy path is much larger. An already deployed contract can expand memory once, then repeatedly pass the *same* slice as `STATICCALL` input to one warm, empty address with zero value and zero forwarded gas. The EVM pays `C(w) = 3w + floor(w²/512)` for `w = ceil(input_len/32)` words once; it does not pay a per-byte copy charge on each subsequent call. In a bounded Foundry fixture, an unrolled `PUSH0/PUSH3/STATICCALL/POP` sequence added 115 gas per warm call (100 for the call, 15 for stack setup and result disposal), and the measured expansion delta matched `C(w)`.

For a `2^24` gas transaction with empty transaction calldata, maximizing `input_len × call_count` in this construction gives roughly **1.69 MB × 96.7k calls = 164 GB of recorded input bytes**, *before* hex/JSON encoding. This is a calculation, not a full-size trace measurement; it includes the 21k intrinsic gas and first cold access, and amortizing jumps in a maximally sized unrolled runtime changes the estimate negligibly. Even a 64 KiB per-frame cap alone still permits about **9.5 GB** over the transaction.

This makes the total byte budget essential. Since both new limits default to unlimited, downstream Reth tracing paths need to set them explicitly; the RPC response should also fail or clearly report truncation, because an ordinary `callTracer` frame has no `full_data_len` field and otherwise looks complete. A bounded regression using repeated calls to the same large slice would exercise the amplification without trying to allocate the projected amount.

christophercampbell:
@mattsse @banteg - PRs [519](https://github.com/paradigmxyz/revm-inspectors/pull/519) & [520](https://github.com/paradigmxyz/revm-inspectors/pull/520) are a better solution than this one. Closing as redundant, thanks!

